Issue #181, opened 2014-12-15
TCK: SubscriberValidation should ship with default Publisher instead of forcing to impl HelperPublisher


ktoso:
If I'm an reactive streams implementer, and my library only has Subscribers, I'd rather not be forced into implementing a proper Publisher in order to fulful `createHelperPublisher(long elements)`.

We can and should provide default implementations of VALID "simple number publishers" you can use.

rkuhn:
Implementation has been merged, closing.

PR #186, opened 2014-12-24
+tck #181 provides default helperPublisher for subscriber tests

ktoso:
This is a rather heavy PR, however it removes the need of implementing a (helper) Publisher when I'm only interested in implementing a Subscriber. This is done by providing a default implementation - based on the example implementations (reused).

The TCK automatically picks the default publisher if it is able to. If not (types don't match), it fails ASAP and tells the implementer that he/she has to implement a custom publisher (like we, until now, always forced implementers to).

Let me know what you think!

```
+ When subscriber verifications are extended using Integer
  we can use the existing publisher implementations from examples
  to drive the tests. This allows implementers who only care about
  implementing Subscribers to not care at all about implementing
  Publishers
+ In case the subscriber verification is extended using a custom signal
  type (defined as "not Integer") we do not magically generate a
  Publisher of this type, but instead fail the tests telling the
  implementer that he/she will need to implement a custom helperPublisher
  which is exactly what we have up until now always demanded from
  implementations.
+ We also provide tests to verify the test failures work as expected,
  and that the messages are indeed helpful.
+ Updated TCK README.md to clarify when to implement a custom publisher
+ fixed mistakenly added `<T>` on NumberPublisher - it is bound to Ints
+ TCK now depends on examples, as it uses the well documented async
  publisher to provide the default publisher for numbers. As well as
  in order to allow implementers to reuse those publishers when
  implementing custom Publishers to drive their Subscriber specs.
```

Refs #181
Merry :christmas_tree: ;-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 language: java
 script:
-- ./gradlew check
+  - ./gradlew check
 cache:
   directories:
   - $HOME/.gradle
 jdk:
-- openjdk6
+  - openjdk6
 env:
   global:
   - TERM=dumb
--- a/examples/src/main/java/org/reactivestreams/example/unicast/AsyncSubscriber.java
+++ b/examples/src/main/java/org/reactivestreams/example/unicast/AsyncSubscriber.java
@@ -1,6 +1,5 @@
 package org.reactivestreams.example.unicast;
 
-import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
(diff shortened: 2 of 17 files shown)


viktorklang:
Superseded by #192 (rebased on top of current master).

ktoso:
rebasing and reopening to verify if validation would/will fail.

ktoso:
Oh... no re-opening powers... :-(

PR #189, opened 2014-12-29
Removes the use of annotations, switches to test prefix naming

### LINKED ISSUE #188: Spec. 2. 12 TCK : Testing a unicast publisher
Hi there,
After updating to the latest 1.0.0.RC1 (#84), Reactor had an issue with spec 2.12 that can be easily fixed but that asks for a quick confirmation/question.

Spec 2.12 forbids double subscription of a given Subscriber to a same Publisher. However we acknowledged there are two kinds of publisher, cold and hot streams. In the cold stream case it seems enough and more efficient to not maintain a registry of subscriptions because each subscription can usually track alone their state: e.g. an Iterable Publisher maintains a single reference to the iterable and a cursor by subscription to track the requests. In our test that is the case where we publish from a helper IterableStream (Streams.from([1,2,3,...]).

However the 2.12 test fails since the two subscriber are actually orthogonal, they start iterating onNext from index 0 and thats pretty much it. To force a shared registry, we offer a Stream#broadcast operation that transforms the cold iterable stream into a hot stream and therefore maintains an active registry of subscriptions (fulfilling the 2.12 contract) that will receive elements T when they are actually subscribed, whatever the cursor upstream from the iterable publisher is actually incremented to.

So if we validate the assumption there are 2 types of streams, should that test mandate use of an hot stream (and therefore requires an hot stream publisher factory method specifically for that test ?). The test should actually validate the cold stream behavior too in fact. That would also need a quick amend to the spec to mention these two specific behaviors: exclusive on hot stream and inclusive on cold streams. 
Another way to look at this is to mandate that the subscriber does actually support only 1 subscription and will assume the ownership of the relation with the publisher, that would actually move the current test to the subscriber verification instead of publisher.
Thoughts ?

viktorklang:
Introduces interfaces to specify which tests should be implemented for the different *Verification classes, which makes it much easier to keep track that test cases have been properly implemented.

*IMPORTANT* Fixes #188 by making the 2.12 rule "untested" in the TCK as it is not reasonably testable.

Removes the annotations that were used for TCK documentation, since they were confusing and not up to date. Replaces them with prefix naming. ("optional"/"required"/"untested"/"stochastic")


viktorklang:
No objections, merging :)

PR #193, opened 2015-01-02
+TCK: Provide helper publisher (rebased)

ktoso:
This is a rebase of: https://github.com/reactive-streams/reactive-streams/pull/186
With the additional commit of making travis failures actually useful - we need `--info` there, otherwise the build log won't include any stacktraces of errors.

Locally it works fine for me, so need to see what makes it fail in Viktor's rebase-PR over here: https://github.com/reactive-streams/reactive-streams/pull/192


ktoso:
Yes, we have trouble - here's the failures:

```
Gradle test > org.reactivestreams.example.unicast.IterablePublisherTest.required_spec317_mustSignalOnErrorWhenPendingAboveLongMaxValue FAILED
    java.lang.AssertionError: Expected onError(java.lang.IllegalStateException) within 100 ms
        Caused by:
        java.lang.AssertionError: Expected onError(java.lang.IllegalStateException) within 100 ms
```

```
Gradle test > org.reactivestreams.example.unicast.IterablePublisherTest.stochastic_spec103_mustSignalOnMethodsSequentially FAILED
    java.lang.AssertionError: Async error during test execution: org.reactivestreams.tck.PublisherVerification$25$1@3232bf23 violated the Reactive Streams rule 3.17 by demanding more elements than Long.MAX_VALUE.
        Caused by:
        java.lang.IllegalStateException: org.reactivestreams.tck.PublisherVerification$25$1@3232bf23 violated the Reactive Streams rule 3.17 by demanding more elements than Long.MAX_VALUE.
Gradle test > org.reactivestreams.example.unicast.IterablePublisherTest.untes
```

```

Gradle test > org.reactivestreams.example.unicast.UnboundedIntegerIncrementPublisherTest.required_spec306_afterSubscriptionIsCancelledRequestMustBeNops STANDARD_ERROR
    java.lang.IllegalStateException: org.reactivestreams.tck.PublisherVerification$15$2@257b746 violated the Reactive Streams rule 2.13 by throwing an exception from onNext or onComplete.
    	at org.reactivestreams.example.unicast.AsyncIterablePublisher$SubscriptionImpl.doSend(AsyncIterablePublisher.java:168)
    	at org.reactivestreams.example.unicast.AsyncIterablePublisher$SubscriptionImpl.doRequest(AsyncIterablePublisher.java:88)
    	at org.reactivestreams.example.unicast.AsyncIterablePublisher$SubscriptionImpl.run(AsyncIterablePublisher.java:197)
    	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1146)
    	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
    	at java.lang.Thread.run(Thread.java:701)
    Caused by: org.reactivestreams.tck.support.SubscriberBufferOverflowException: Received more than bufferSize (32) onNext signals. The Publisher probably emited more signals than expected!
    	at org.reactivestreams.tck.TestEnvironment$ManualSubscriber.onNext(TestEnvironment.java:236)
    	at org.reactivestreams.tck.TestEnvironment$ManualSubscriberWithSubscriptionSupport.onNext(TestEnvironment.java:442)
    	at org.reactivestreams.tck.PublisherVerification$15$2.onNext(PublisherVerification.java:600)
    	at org.reactivestreams.example.unicast.AsyncIterablePublisher$SubscriptionImpl.doSend(AsyncIterablePublisher.java:154)
    	... 5 more
    Caused by: java.lang.IllegalStateException: Queue full
    	at java.util.AbstractQueue.add(AbstractQueue.java:99)
    	at java.util.concurrent.ArrayBlockingQueue.add(ArrayBlockingQueue.java:237)
    	at org.reactivestreams.tck.TestEnvironment$Receptacle.add(TestEnvironment.java:791)
    	at org.reactivestreams.tck.TestEnvironment$ManualSubscriber.onNext(TestEnvironment.java:233)
    	... 8 more
```

I think they all originate back to the publisher now being async, need to look into this in-depth (setting expectations here though - in-depth I'll probably be only able to do from the 4th Jan on).

@viktorklang

viktorklang:
>Gradle test > org.reactivestreams.example.unicast.IterablePublisherTest.required_spec317_mustSignalOnErrorWhenPendingAboveLongMaxValue FAILED
    java.lang.AssertionError: Expected onError(java.lang.IllegalStateException) within 100 ms
        Caused by:
        java.lang.AssertionError: Expected onError(java.lang.IllegalStateException) within 100 ms


Isn't this just a case of TravisCI being slower than our local machines?

viktorklang:
@ktoso and the next failure is just due to the first failure, since that ABQ is just 1 element big?

ktoso:
Brain stormed with @viktorklang and fixed the "timeout is too little for travis" problem by introducing env overrides for all timeouts. They also got default values now, please check https://github.com/ktoso/reactive-streams/commit/8769da9c453cffcee275804c3e0f7741d2492425 for changes around that. In general it's now possible to `export DEFAULT_TIMEOUT_MILLIS=300` and the TCK will use it.

Tests passed on Travis now with increased timeouts.
We can still see the failure from the async publisher in `required_spec306_afterSubscriptionIsCancelledRequestMustBeNops` but it's not failing the test, need to look deeper into that one once I get off down from the moutnains :-)

viktorklang:
Since it is the following test that fails:

~~~java
@Test // Must be here for TestNG to find and run this, do not remove
public class UnboundedIntegerIncrementPublisherTest extends PublisherVerification<Integer> {

  private ExecutorService e;
  @BeforeClass void before() { e = Executors.newFixedThreadPool(4); }
  @AfterClass void after() { if (e != null) e.shutdown(); }

  public UnboundedIntegerIncrementPublisherTest() {
    super(new TestEnvironment());
  }

  @Override public Publisher<Integer> createPublisher(long elements) {
    return new InfiniteIncrementNumberPublisher(e);
  }

  @Override public Publisher<Integer> createErrorStatePublisher() {
    return null;
  }

  @Override public long maxElementsFromPublisher() {
    return super.publisherUnableToSignalOnComplete();
  }
}
~~~

I'd suggest you have another look at how `publisherUnableToSignalOnComplete` affects the PublisherVerification, I suspect there's a missing check somewhere and then we request a huge number (I increased the ABQ size to 32k and it still overflows).

ktoso:
@viktorklang latest commit fixes a "leaked still alive and kicking publisher" which leaked from `spec303`, yet caused the stacktrace we noticed in `required_spec306_afterSubscriptionIsCancelledRequestMustBeNops` (since it's async, and keeps publishing, it accidentally blows up during the execution of spec306 - which is run after the 303 spec).

The new impl is much safer in face of different Publishers, please review https://github.com/ktoso/reactive-streams/commit/8ed7a4189036d285fa264fddcb98b9701e2fe227 - the commit message has more details and there's some comments in line explaining rationale on skipping `super.onNext()` for example.

viktorklang:
Awesome that you found the issue, @ktoso!

ktoso:
Addressed feedback, simplified code a bit - no need for additional configuration method, we're using `final long oneMoreThanBoundedLimit = boundedDepthOfOnNextAndRequestRecursion() + 1;` as "ok, enough elements signalled, I can stop the test already"

ktoso:
refs https://github.com/reactive-streams/reactive-streams/issues/181

examples/src/test/java/org/reactivestreams/example/unicast/AsyncSubscriberTest.java
```diff
@@ -61,7 +52,15 @@ public AsyncSubscriberTest() {
...
-    latch.await(DefaultTimeoutMillis * 10, TimeUnit.MILLISECONDS);
+    latch.await(env.defaultTimeoutMillis() * 10, TimeUnit.MILLISECONDS);
     assertEquals(i.get(), 45);
   }
+
+  @Override public Integer createElement(int element) {
+    return element;
+  }
+
+  @Override public Publisher<Integer> createHelperPublisher(long elements) {
```
viktorklang: This override seems meaningless :)

tck/README.md
```diff
@@ -51,7 +51,7 @@ Specification rule abides the following naming convention: `spec###_DESC` where:
 ```
 
 The prefixes of the names of the test methods are used in order to signify the character of the test. For example, these are the kinds of prefixes you may find:
-"required_", "optional_", "stochastic_", "untested_".
+"`required_`", "`optional_`", "`stochastic_`", "`untested_`".
```
viktorklang: Nice!

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
```diff
@@ -109,15 +125,24 @@ public boolean skipStochasticTests() {
...
-   *
-   * @param elements exact number of elements this publisher should emit,
-   *                 unless equal to {@code Long.MAX_VALUE} in which case the stream should be effectively infinite
+   * Helper method required for creating the Publisher to which the tested Subscriber will be subscribed and tested against.
+   * <p>
+   * By default an <b>asynchronously signalling Publisher</b> is provided, which will use
+   * {@link org.reactivestreams.tck.SubscriberBlackboxVerification#createElement(int)} to generate elements type
+   * your Subscriber is able to consume.
+   * <p>
+   * Sometimes you may want to implement your own custom custom helper Publisher - to validate behaviour of a Subscriber
```
viktorklang: A recommendation is to apply the TCK to your test publisher :)

viktorklang:
:+1:, this looks excellent Konrad!

@reactive-streams/contributors please vote for this TCK improvement before monday the 12th of January, or ask for more time to think about this before the 12th. Thanks!

examples/src/test/java/org/reactivestreams/example/unicast/AsyncSubscriberTest.java
ktoso: thanks, removed!

tck/README.md
ktoso: (:

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
ktoso: In fact, this method should NOT be here! Since we're extending in WithHelperPublisher, also noticed a javadoc `{@link}` error in there, fixed as well.

ktoso: Always good to re-reread a PR... Thanks for reviewing!

.travis.yml
```diff
@@ -1,11 +1,13 @@
 language: java
 script:
-- ./gradlew check
+  - ./gradlew check --info
```
viktorklang: Should this be left in?

tck/README.md
```diff
@@ -286,13 +350,41 @@ public class MySubscriberWhiteboxVerificationTest extends SubscriberWhiteboxVeri
...
+
+  public RangePublisherTest() {
+    super(new MySubscriberTest(DEFAULT_TIMEOUT_MILLIS));
   }
 
+  // ...
 }
 ```
 
+Note that hard-coded values *take precedence* over environment set values (!).
```
viktorklang: Perhaps better to word it as "Note that explicitly passed in values take precedence over values provided by the environment"

.travis.yml
ktoso: I think so, yes - because [+travis run gradle with --info; otherwise we don't get any meaningful stacktraces out of the travis test runs](https://github.com/ktoso/reactive-streams/commit/00c7a51320fd6d493e107502f68f46b72df618ca). Without it we get failures on travis which are as helpful as "there was a failure", with `--info` we get a stacktrace of the failure.

viktorklang:
Aside from comments, LGTM!

.travis.yml
viktorklang: Fair enough :)

tck/README.md
ktoso: yes, that's better - thanks!

.travis.yml
alkemist: `--info` adds a lot of noise.

Add this to about line 27 of build.gradle…

```
  tasks.withType(Test) {
    testLogging {
      exceptionFormat "full"
      events "failed", "started"
    }
  }
```

ktoso: Nice, thanks @alkemist! I'll include it in this PR then
`started` makes more sense than `passed` for us I guess, since we care about the skipped ones being run as well.

tck/README.md
ktoso: // github is acting weird, I updated that line yet this comment remains expanded - https://github.com/reactive-streams/reactive-streams/pull/193/files#diff-4ce068920e11452416705ccb8f46ec78R205

viktorklang:
Alright, LGTM, merging this since review feedback has been addressed.

PR #199, opened 2015-01-12
+tck #198 allow 0-length publishers to be tested

### LINKED ISSUE #198: TCK PublisherVerification does not support publishers of empty streams
This code in the `PublisherVerification` class:

```java
  @Override @Test
  public void required_validate_maxElementsFromPublisher() throws Exception {
    assertTrue(maxElementsFromPublisher() > 0, "maxElementsFromPublisher MUST return a number > 0");
  }
```

Causes an `EmptyPublisher` class representing an empty stream to fail conformance testing.

An empty stream may be an edge case but it is a legitimate stream, and a Reactive Streams implementation should be able to have such a `Publisher`. One use case would be testing that `Subscriber`s operate correctly when subscribed to empty streams.

Therefore, TCK should allow a subclass of `PublisherVerification` to return zero from `maxElemensFromPublisher()`, and if it does TCK should test that `request()`-ing any elements at all from the `Publisher` results in only a call to the `Subscriber`'s `onComplete()`.

ktoso:
Adapts the TCK to allow "0 length publishers", though only one test makes sense for them AFAICS.

This PR resolves #198 and adheres to current reactive streams rules about "it must fist give out an subscription" - as seen in the call-order-schema on https://github.com/reactive-streams/reactive-streams#api-components
Though there is an open question (from my end) about if this should be explicitly mentioned in the spec itself. (See #198 for a discussion).

### DIFF
--- a/tck/src/test/java/org/reactivestreams/tck/EmptyLazyPublisherTest.java
+++ b/tck/src/test/java/org/reactivestreams/tck/EmptyLazyPublisherTest.java
@@ -1,0 +1,42 @@
+package org.reactivestreams.tck;
+
+import org.reactivestreams.example.unicast.AsyncIterablePublisher;
+import org.reactivestreams.Publisher;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import java.util.Collections;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+@Test
+public class EmptyLazyPublisherTest extends PublisherVerification<Integer> {
+
+  private ExecutorService ex;
+
+  public EmptyLazyPublisherTest() {
+    super(new TestEnvironment());
+  }
+
+  @BeforeClass
+  void before() { ex = Executors.newFixedThreadPool(4); }
+
+  @AfterClass
+  void after() { if (ex != null) ex.shutdown(); }
+
+  @Override
+  public Publisher<Integer> createPublisher(long elements) {
+    return new AsyncIterablePublisher<Integer>(Collections.<Integer>emptyList(), ex);
+  }
+
+  @Override
+  public Publisher<Integer> createErrorStatePublisher() {
+    return null;
+  }
+
+  @Override
+  public long maxElementsFromPublisher() {
+    return 0;
+  }
+}
--- a/tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
+++ b/tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
@@ -185,6 +185,35 @@ public class PublisherVerificationTest extends TCKVerificationSupport {
   }
 
   @Test
+  public void optional_spec105_emptyStreamMustTerminateBySignallingOnComplete_shouldNotAllowEagerOnComplete() throws Throwable {
+    final Publisher<Integer> illegalEmptyEagerOnCompletePublisher = new Publisher<Integer>() {
+      @Override public void subscribe(final Subscriber<? super Integer> s) {
+        s.onComplete();
+      }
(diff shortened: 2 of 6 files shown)


examples/src/test/java/org/reactivestreams/example/unicast/AsyncSubscriberTest.java
```diff
@@ -31,7 +30,6 @@ public AsyncSubscriberTest() {
 
   @Override public Subscriber<Integer> createSubscriber() {
     return new AsyncSubscriber<Integer>(e) {
-      private long acc;
```
ktoso: boyscouting here - this was never used // cc @viktorklang

gnp:
In the name of this issue *and* in the patch, "length" is mis-spelled "lenght".

ktoso:
Thanks a lot for catching this! Fixed now.

viktorklang:
@reactive-streams/contributors Please review/comment/vote before the 16th of January, don't hesitate to ask for more time if you are currently unavailable to review within the time frame.

NOTE: this is not about changing spec semantics, only to make sure the TCK can cope with empty publishers.

tck/src/test/java/org/reactivestreams/tck/EmptyLazyPublisherTest.java
```diff
@@ -0,0 +1,45 @@
...
+@Test
+public class EmptyLazyPublisherTest extends PublisherVerification<Integer> {
+
+  public EmptyLazyPublisherTest() {
+    // TODO these params will be removed once https://github.com/reactive-streams/reactive-streams/pull/193 is merged
+    super(new TestEnvironment(300), 500);
+  }
+
+  @Override
+  public Publisher<Integer> createPublisher(long elements) {
```
viktorklang: Isn't it easier to reuse `AsyncIterablePublisher` with a Collections.emptyList?

tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
```diff
@@ -188,6 +188,51 @@ public void required_spec105_mustSignalOnCompleteWhenFiniteStreamTerminates_shou
...
+        throw new TestException();
+      }
+    };
+
+    requireTestSkip(new ThrowingRunnable() {
+      @Override
+      public void run() throws Throwable {
+        customPublisherVerification(failingPublisher).optional_spec105_emptyStreamMustTerminateBySignallingOnComplete();
+      }
+    }, "This test is specific for 0-length streams");
```
viktorklang: Perhaps I'm daft today, but I don't understand this test case. Why does it throw an exception in subscribe?

viktorklang:
Aside from comments, LGTM

tck/src/test/java/org/reactivestreams/tck/EmptyLazyPublisherTest.java
ktoso: good point, thanks!

tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
ktoso: This test-of-test was meant to showcase we skip this test if the publisher is not a 0-length one.
However this skipping is not really needed now that I had the time to think about it again.
Removed this and keeping the test enabled for all publishers which can signal completion.

ktoso:
Updated according to comments; the new 0-length test is `optional` as some publishers will not be able to signal 0 elements (any publisher that "I must emit my one element and then complete" - like FuturePublisher) and we don't want implementers to special case this 0-lenght case.

rkuhn:
LGTM

viktorklang:
LGTM! Merging without complaints :)

PR #205, opened 2015-01-21
Fixes javadoc issues to make the build Java8 compliant

viktorklang:
@reactive-streams/contributors This is just a bugfix PR to make RS javadoc succeed on JDK8, will merge when TravisCI clears it.


viktorklang:
Mergin'

PR #207, opened 2015-01-29
make build.gradle support idea and eclipse

darionyaphet:
add some lines on build.gradle to support idea and eclipse

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,8 @@
 subprojects {
     apply plugin: "java"
     apply plugin: "osgi"
+    apply plugin: "idea"
+    apply plugin: "eclipse"
 
     group = "org.reactivestreams"
     version = "1.0.0.RC1"


viktorklang:
Hi @darionyaphet,

Welcome to Reactive Streams!

We discussed this a while back and the consensus was to put those plugins in your user/global gradle config. Does that work for you?

Cheers,
√

darionyaphet:
hi cherrs 
Actually I don't know how to use user/global gradle config . 
I reading Kafka's gradle config file , Kafka put this config into build.gradle .
And what is difference put it into diff config-file ? thank you for you help :)

viktorklang:
Does this help?
https://stackoverflow.com/questions/11567786/global-plugins-in-gradle

darionyaphet:
yes ! 
This support is very useful and many project will use it . 
Put it into global config maybe better :+1:

viktorklang:
Fantastic!

I'll close this PR then :)

Cheers,
√

PR #208, opened 2015-01-29
Repairs formatting issue of tables in spec README

### LINKED ISSUE #206: github fails to render tables in readme.md spec
Just noticed that github stopped properly rendering the table layout we used for the spec:

reactive-streams_reactive-streams

Not sure since when exactly. It seems though like github changed something in their rendering, we didn't touch that layout for a loong time.

ktoso:
Resolves #206

The problem was introduced in https://github.com/reactive-streams/reactive-streams/commit/a3005b779344bf8ed6047d73d89ad9b39bb1d1a7 in which an ending `.` was added to each spec rule. By mistake it was also added to the table header, which broke the rendering :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,8 +76,8 @@ public interface Publisher<T> {
 }
 ````
 
-| ID                        | Rule                                                                                                  . |
-| ------------------------- | ------------------------------------------------------------------------------------------------------. |
+| ID                        | Rule                                                                                                   |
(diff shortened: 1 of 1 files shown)


viktorklang:
Thanks for the diligence, @ktoso!

@reactive-streams/contributors Merging this cosmetic fix

PR #214, opened 2015-02-10
Fixes #210 by removing 1.12 and repurposing its TCK checks for 1.09

### LINKED ISSUE #210: 1.12 is a duplicate of 1.9
"1.9	Calling Publisher.subscribe MUST return normally except when the provided Subscriber is null in which case it MUST throw a java.lang.NullPointerException to the caller, for all other situations the only legal way to signal failure (or reject a Subscriber) is via the onError method."

"1.12	A Publisher MAY reject calls to its subscribe method if it is unable or unwilling to serve them [1]. If rejecting it MUST do this by calling onError on the Subscriber passed to Publisher.subscribe instead of calling onSubscribe."

We can, without impacting the semantics of the spec, remove rule 12 and rename 1.13 to 1.12.

viktorklang:
@reactive-streams/contributors This PR does not semantically alter the spec, it just collapses 1.12 into 1.09, making the spec simpler. Please object to this change before the 13th of February if at all, thank you!


benjchristensen:
:+1: as per https://github.com/reactive-streams/reactive-streams/issues/210#issuecomment-73802402

viktorklang:
@reactive-streams/contributors Merging without objection :)

PR #219, opened 2015-02-15
=tck minor test name fixup, it is a required test

ktoso:
Trivial name fixup, introduced in https://github.com/reactive-streams/reactive-streams/pull/214/files#diff-4ebca955800e47695d157fcd89b4fec6R263 
Notice that this rule is a required rule, thus the method name of the test should reflect this.

( see why such name here: https://github.com/reactive-streams/reactive-streams/pull/219/files#diff-4ebca955800e47695d157fcd89b4fec6R273 )

### DIFF
--- a/tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
+++ b/tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
@@ -264,7 +264,7 @@ public class PublisherVerificationTest extends TCKVerificationSupport {
   }
 
   @Test
-  public void optional_spec109_mayRejectCallsToSubscribeIfPublisherIsUnableOrUnwillingToServeThemRejectionMustTriggerOnErrorAfterOnSubscribe_actuallyPass() throws Throwable {
+  public void required_spec109_mayRejectCallsToSubscribeIfPublisherIsUnableOrUnwillingToServeThemRejectionMustTriggerOnErrorAfterOnSubscribe_actuallyPass() throws Throwable {
     customPublisherVerification(SKIP, new Publisher<Integer>() {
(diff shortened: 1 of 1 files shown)


viktorklang:
Nice catch, LGTM

tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
```diff
@@ -264,7 +264,7 @@ public void required_spec109_subscribeThrowNPEOnNullSubscriber_shouldFailIfDoesn
   }
 
   @Test
-  public void optional_spec109_mayRejectCallsToSubscribeIfPublisherIsUnableOrUnwillingToServeThemRejectionMustTriggerOnErrorAfterOnSubscribe_actuallyPass() throws Throwable {
+  public void required_spec109_mayRejectCallsToSubscribeIfPublisherIsUnableOrUnwillingToServeThemRejectionMustTriggerOnErrorAfterOnSubscribe_actuallyPass() throws Throwable {
     customPublisherVerification(SKIP, new Publisher<Integer>() {
```
viktorklang: @ktoso Why is it `SKIP` here tho?

ktoso: That's OK, because this test is for the error state publisher. `customPublisherVerification` first the normal publisher, then the error state publisher - we're testing the error state publisher here, thus only the 2nd param.

// In the test we have `whenHasErrorPublisherTest(...)`

viktorklang:
@smaldini We should make sure we get this into the RC3 as well

rkuhn:
This was slated for RC3, merging for RC4 instead.

PR #220, opened 2015-02-16
read signing parameters from the console

rkuhn:
Because putting the passphrase in a file is just plain unacceptable and
putting it on the command line makes it visible to everyone else on the
machine in addition (depending on the environment). The code snippet has
just been copied verbatim from
http://gradle.org/docs/current/userguide/signing_plugin.html.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -136,7 +136,7 @@ subprojects {
 import org.gradle.plugins.signing.Sign
 
 gradle.taskGraph.whenReady { taskGraph ->
-    if (taskGraph.allTasks.any { it instanceof Sign }) {
+    if (taskGraph.allTasks.any { it instanceof Sign } && signing.required) {
         // Use Java 6's console to read from the console (no good for
         // a CI environment)
         Console console = System.console()
--- a/build.gradle
+++ b/build.gradle
@@ -132,3 +132,26 @@ subprojects {
         uploadArchives.enabled = false
     }
 }
+
+import org.gradle.plugins.signing.Sign
+
+gradle.taskGraph.whenReady { taskGraph ->
+    if (taskGraph.allTasks.any { it instanceof Sign }) {
+        // Use Java 6's console to read from the console (no good for
+        // a CI environment)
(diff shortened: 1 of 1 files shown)


viktorklang:
Test failed?

smaldini:
Might need to add an environment detection.

rkuhn:
Is travis configured to sign things? (I’m far from an expert but this change should only apply to tasks that involve some signing)

viktorklang:
Seems like this change should be modified to only apply for the upload task?

rkuhn:
But which other task uses signing? The console input is only asked for if at least one task wants to do some signing, which I verified locally (`./gradlew test` does not prompt).

viktorklang:
@rkuhn Perhaps Travis adds its own tasks that uses signing?

benjchristensen:
If you need to do signing I suggest looking at using Travis+Bintray which can encrypt keys into the file stored in Github. For example, RxJava uses this so all releases can be done via publicly accessible config in Github: 

- https://github.com/ReactiveX/RxJava/blob/1.x/.travis.yml#L12
- https://github.com/ReactiveX/RxJava/blob/1.x/gradle/buildViaTravis.sh#L12

Then the Maven Central signing is done via a key stored in Bintray, while the Bintray and Sonatype credentials are encrypted into the .travis.yml file. 

Documentation for Travis can be found here: http://docs.travis-ci.com/user/encryption-keys/

The great thing about this approach is that the entire process is then automated and can be run by anyone with commit privileges to the repo, just by committing a tag (which can be done right from the Github Releases UI). 

Examples of what the build looks like in Travis:

- https://travis-ci.org/ReactiveX/RxJava/builds/53416482#L91
- https://travis-ci.org/ReactiveX/RxJava/builds/53416482#L416
- https://travis-ci.org/ReactiveX/RxJava/builds/53416482#L3375

viktorklang:
This sounds like a great idea, thanks for the heads up, Ben

rkuhn:
giving up

PR #221, opened 2015-02-16
Removes superfluous creation of processors for SubscriberWhiteboxVerific...

viktorklang:
...ation

@smaldini does this address your issue?


smaldini:
We will consider a new issue regarding why the CI passes where it shouldn't it which involves adding processor examples, but I suggest to go with #222 if it passes approval now and complete in the next candidat. I'd very much like to release before holiday a stable version :D That would be an awesome present.

viktorklang:
Ok, closing this w/o merging!

PR #225, opened 2015-02-17
remove rule 1:12 (produce same elements to all Subscribers)

### LINKED ISSUE #224: remove rule 1:12 (must produce same elements to multiple Subscribers)
Rationale: rule 1:11 states that a Publisher may choose to treat multiple Subscribers as either unicast or multicast recipients, but 1:12 forbids the delivery of different stream elements to different subscribers, thereby effectively making the unicast choice useless—the typical use-case would be to have a Publisher set up a new connection to some data source for every new Subscriber, and then it cannot in general guarantee the “same elements” restriction (which might also be unwanted).

Removal of the rule does not actually change the set of permissible behaviors apart from removing the aforementioned conflict, since the word “multicast” sufficiently captures the intent that was behind 1:12.

rkuhn:
This rule is in conflict with 1:11 which allows a Publisher to treat
multiple Subscribers as either as unicast or multicast recipients. The
verification of proper multicast behavior (which 1:12 specified) has
been retained, the test methods renamed accordingly.

Fixes #224.


tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
```diff
@@ -635,7 +635,7 @@ public void run(Publisher<T> pub) throws Throwable {
 
   // Verifies rule: https://github.com/reactive-streams/reactive-streams#1.12
```
ktoso: outdated link to 1.12 here

rkuhn:
There was agreement on the ticket, no comments here, so merging after 23 days.

Issue #236, opened 2015-03-03
Clarification on Whitebox subscriber verification


purplefox:
I am attempting to implement Whitebox subscriber verification as explained in the docs, but I am unsure what I supposed to do in the createSubscriber() method.

I understand that this should return an instance of my subscriber class, but with the methods overridden to call the probe, but I do not understand what "in addition to normal Subscriber work that you're testing," means.

Does this mean I should call the super class method first before calling the probe, or perhaps I should call the super class method after I call the probe, or perhaps something else?

Any clarification on what is expected by the TCK here would be greatly appreciated :)

ktoso:
> [...] what "in addition to normal Subscriber work that you're testing," means.

Since the TCK is aimed at testing real implementations, they'll probably be doing something with the received data (emit it somewhere, write it to a file, drop it, ...), this is what is meant by "normal subscribe work". Basically it boils down to: You should not write a subscriber just for the sake of "I have a subscriber which passes the TCK", and then have completely different ones do the actual work in your library :-)

If you can, you can adapt an existing implementation by subclassing - calling the probe methods where needed (see tck/README.md for an example) and call super so that it actually performs the work you're testing.

purplefox:
Thanks,

I got that I should subclass my Subscriber method by looking at the example in the README.

However that example does not call super class implementation, and I can't infer from that whether the probe should be called before or after the call to the super class.

purplefox:
(I think it would be helpful to edit the example in the README so it actually calls the super class methods)

ktoso:
I'm not sure we will be able to give an universal recipe for this... Feels like the exact "where" are implementation specific.

They may be doing weird async things in there etc. 
When updating the README https://github.com/reactive-streams/reactive-streams-jvm/issues/237 I'll try to think of a better example, but my gut feeling is there isn't an one-size-fits-all solution here.

purplefox:
I don't think we need a universal recipe here.

Just want to know what I should be doing in the case that the Subscriber class is easily subclassable, like  ours is (and many others will be).

ktoso:
Yeap, agree on that one.

viktorklang:
Can this be closed or does something need to be asnwered/fixed?

ktoso:
Not yet, the PR fixing this precise question is not in yet - hoping to get it in today. 
It's about prepping an example of using whitebox in the simplest case.

viktorklang:
Awesome, thanks Konrad

[...]

PR #209, opened 2015-02-04
Modifies rules 1.09 and 2.13 to mandate `java.lang.NullPointerException`...

viktorklang:
... be thrown.

Updates the TCK, Spec and example implementations.

@reactive-streams/contributors As discussed in #204, here is my proposed spec, tck, example amendment for `null` handling. Please comment and vote **before the 9th of February 2015** as we should attempt to ship 1.0.0.RC2 STAT.

PR #244, opened 2015-03-14
+tck #236 example subscriber whitebox tested, and whitebox fixed

ktoso:
Adds better example on using `SubscriberWhiteboxVerification` - it now explains how you would use it with your impl, thanks @purplefox for pointing out the need for clarification here in #236!

This PR also makes us use the whitebox to test the example subscriber. This found a bug in the new tests introduced during https://github.com/reactive-streams/reactive-streams-jvm/pull/209/files as it accidentally subscribed twice (whitebox's Stage is subscribing the `sub` to the `pub` during creation). Fixed tests, and improved docs on these methods.

Resolves https://github.com/reactive-streams/reactive-streams-jvm/issues/236


ktoso:
// cc @viktorklang

tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
```diff
@@ -560,6 +553,10 @@ public SubscriberPuppet puppet() {
 
     public T signalNext() throws InterruptedException {
       T element = nextT();
+      return signalNext(element);
+    }
+
+    public T signalNext(T element) throws InterruptedException {
```
viktorklang: Public?

tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
```diff
@@ -560,6 +553,10 @@ public SubscriberPuppet puppet() {
 
     public T signalNext() throws InterruptedException {
       T element = nextT();
+      return signalNext(element);
```
viktorklang: `return signalNext(nextT())`?

viktorklang:
Aside from comments, :+1:!

tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
ktoso: can be private, yeap

tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
ktoso: sure

ktoso:
Comments addresses :+1:

viktorklang:
@reactive-streams/contributors this is a documentation fix and bugfix for the TCK, will merge in 48h if no objections have been made

benjchristensen:
Since this is all in the TCK I'm assuming it's fixing conformance with the spec, so I'm okay with it.

PR #241, opened 2015-03-11
!tck - clarifies error publisher, and method name change (?)

ktoso:
+ add better readme on what this method is
+ add better javadoc on this method
- removes reference to old style spec annotation from readme
+ proposing to change method name to "createFailed..." as it is the wording used in the spec and reactive manifesto (footnote 1.1)

Resolves #237, #235

PR #246, opened 2015-03-20
=tck general tck/readme.md cleanup so it matches current code / spec

### LINKED ISSUE #99: TCK Upgrade Story
This issue came up during discussions on the "bring back the TCK" pull request: https://github.com/reactive-streams/reactive-streams/pull/91

In general we need some story about when we change the spec, and the tests, that implementors will be notified about some changed (perhaps *optional*!) behaviour. In the TCK tests which are optional are skipped if they fail to verify currently, this could cause adding optional behaviour in `0.x+1` simply be skipped to be verified, and the implementor may assume "oh good, my impl covers everything". This is not really limited to adding optional tests, but in general behaviour which we can't test for example too.

This issue means to start a discussion about if we want to provide more info for implementors in the TCK itself or simply assume implementors will read changelogs and apply all changes.

One idea on the more extreme side would be that each spec would contain a value `conformsTo() { return Spec_0_4; }`), then if the dependency is updated to `0.5`, we fail the tests listing what has changed... This might be overkill, and simply "please read the change logs" could be our solution here.

I'm interested if you guys see this as a problem that needs solving or we simply document that "if you upgrade, please do read the changelog in detail".

ktoso:
Readme updates so that it shows up to date code as well as explains the TCK upgrade story.

Resolves #99
Depends on #241

:airplane:

### DIFF
--- a/tck/README.md
+++ b/tck/README.md
@@ -87,10 +87,8 @@ Usually this means that this test case can yield false positives ("be green") ev
 
 <a name="type-untested"></a>
 The `untested_` means that the test case is not implemented, either because it is inherently hard to verify (e.g. Rules which use
-the wording "*SHOULD consider X as Y*") or have not been implemented yet (though we hope we have implemented all we
-could!). Such tests will show up in your test runs as `SKIPPED`, with a message pointing out that the TCK is unable to
-validate this Rule. We would be delighted if you can figure out a way to deterministically test Rules, which have been
-marked with this prefix – pull requests are very welcome!
+the wording "*SHOULD consider X as Y*"). Such tests will show up in your test runs as `SKIPPED`, with a message pointing out that the TCK is unable to validate this Rule. If you figure out a way to deterministically test Rules which have been
+marked with this prefix – pull requests are encouraged!
 
 ### Test isolation
 
@@ -200,12 +198,11 @@ public class RangePublisherTest extends PublisherVerification<Integer> {
 
 Notable configuration options include:
 
-* `maxElementsFromPublisher` – which should only be overridden in case the Publisher under test is not able to provide
-  arbitrary length streams, e.g. it's wrapping a `Future<T>` and thus can only publish up to 1 element. In such case you
-  should return `1` from this method. It will cause all tests which require more elements in order to validate a certain
+* `maxElementsFromPublisher` – must be overridden in case the Publisher being tested is of bounded length, e.g. it's wrapping a `Future<T>` and thus can only publish up to 1 element, in which case you
+  would return `1` from this method. It will cause all tests which require more elements in order to validate a certain
   Rule to be skipped,
-* `boundedDepthOfOnNextAndRequestRecursion` – which should only be overridden in case of synchronous Publishers.
-  This number will be used to validate if a `Subscription` actually solves the "unbounded recursion" problem (Rule 3.3).
+* `boundedDepthOfOnNextAndRequestRecursion` – which must be overridden when verifying synchronous Publishers.
+  This number returned by this method will be used to validate if a `Subscription` adheres to Rule 3.3 and avoids "unbounded recursion".
 
 ### Timeout configuration
 Publisher tests make use of two kinds of timeouts, one is the `defaultTimeoutMillis` which corresponds to all methods used
@@ -215,7 +212,7 @@ by the Publisher.
 
 In order to configure these timeouts (for example when running on a slow continious integtation machine), you can either:
 
-**Use env variables** to set these timeouts, in which case the you can just:
+**Use env variables** to set these timeouts, in which case the you can do:
 
 ```bash
 export DEFAULT_TIMEOUT_MILLIS=300
@@ -472,7 +469,7 @@ public class MyIdentityProcessorVerificationTest extends IdentityProcessorVerifi
 
   @Override
   public Publisher<Integer> createFailedPublisher() {
-    // return Publisher that just signals onError instead of null to run additional tests
+    // return Publisher that only signals onError instead of null to run additional tests
     // see this methods JavaDocs for more details on how the returned Publisher should work.
     return null;
   }
--- a/tck/README.md
+++ b/tck/README.md
@@ -10,11 +10,11 @@ The TCK is implemented using **plain Java (1.6)** and **TestNG** tests, and shou
 The TCK aims to cover all rules defined in the Specification, however for some rules outlined in the Specification it is
 not possible (or viable) to construct automated tests, thus the TCK does not claim to completely verify an implementation, however it is very helpful and is able to validate the most important rules.
 
-The TCK is split up into 4 files JUnit 4 test classes which should be extended by implementers, providing their `Publisher` / `Subscriber` implementations for the test harness to validate them. The tests are split in the following way:
+The TCK is split up into 4 TestNG test classes which should be extended by implementers, providing their `Publisher` / `Subscriber` implementations for the test harness to validate them. The tests are split in the following way:
 
 * `PublisherVerification`
-* `SubscriberBlackboxVerification`
 * `SubscriberWhiteboxVerification`
+* `SubscriberBlackboxVerification`
 * `IdentityProcessorVerification`
 
 The next sections include examples on how these can be used and describe the various configuration options.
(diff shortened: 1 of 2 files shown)


tck/README.md
```diff
@@ -13,8 +13,8 @@ not possible (or viable) to construct automated tests, thus the TCK does not cla
 The TCK is split up into 4 files JUnit 4 test classes which should be extended by implementers, providing their `Publisher` / `Subscriber` implementations for the test harness to validate them. The tests are split in the following way:
```
viktorklang: JUnit? I thought we used TestNG?

ktoso: oh wow, good catch

tck/README.md
```diff
@@ -38,8 +38,9 @@ Since the TCK is aimed at Reactive Stream implementers, looking into the sources
 and should help during a libraries implementation cycle.
 
 In order to make mapping between test cases and Specification rules easier, each test case covering a specific
-Specification rule abides the following naming convention: `spec###_DESC` where:
+Specification rule abides the following naming convention: `TYPE_spec###_DESC` where:
 
+* `TYPE` is one of: `required`, `optional`, `stochastic` or `untested` which describe if this test is covering a Rule that MUST or SHOULD be implemented. The specific words are explained in detail below.
```
viktorklang: `The specific words are explained in detail below.` <- where below (i.e. it is not quick to find)

viktorklang: How about linking each of the `TYPE`s to the section that explains it?

tck/README.md
```diff
@@ -66,7 +67,7 @@ Explanations:
 @Test public void optional_spec104_mustSignalOnErrorWhenFails() throws Throwable
 ```
 
-... means that this test case is optional, it covers a *MAY* or *SHOULD* Rule of the Specification. This prefix is also used if more configuration is needed in order to run it, e.g. `@Additional(implement = "createErrorStatePublisher") @Test` signals the implementer that in order to include this test case in his test runs, (s)he must implement the `Publisher<T> createErrorStatePublisher()` method.
+... means that this test case is optional, it covers a *MAY* or *SHOULD* Rule of the Specification. This prefix is also used if more configuration is needed in order to run it, e.g. `@Additional(implement = "createFailedPublisher") @Test` signals the implementer that in order to include this test case in his test runs, (s)he must implement the `Publisher<T> createFailedPublisher()` method.
```
viktorklang: remove "his", use "one" iso "(s)he"

tck/README.md
```diff
@@ -456,32 +459,65 @@ Since you inherit these tests instead of defining them yourself it's not possibl
...
+
+  @Override
+  public Integer createElement(int element) {
+    return element;
   }
 
   @Override
   public Processor<Integer, Integer> createIdentityProcessor(int bufferSize) {
-    return /* ... */;
+    return new MyProcessor<Integer, Integer>(); // return your implementation
```
viktorklang: MyProcessor here does not respect the "bufferSize" param, is that intentional?

tck/README.md
```diff
@@ -456,32 +459,65 @@ Since you inherit these tests instead of defining them yourself it's not possibl
...
   }
 
-  @Override // override the test method, and provide a reason on why you're doing so in the notVerified() message
-  public void spec999_mustDoVeryCrazyThings() throws Throwable {
-    notVerified("Unable to implement test because ...");
+  @Override
+  public Publisher<Integer> createFailedPublisher() {
+    return null; // skip error publisher tests
```
viktorklang: It's implied, but I'd rather have it explicit that: // returning null means that the error tests for the publisher will be skipped

tck/README.md
```diff
@@ -456,32 +459,65 @@ Since you inherit these tests instead of defining them yourself it's not possibl
...
   }
 
 }
 ```
 
-## Upgrade story
+## Upgrading the TCK to newer versions
+While we do not expect the Reactive Streams specification to change in the forseeable future,
+it may happen that some semantics may need changing in the future. In this case you should expect test
```
viktorklang: Multiple uses of "future". Use neutral language iso "we"

tck/README.md
ktoso: +1, they ful explanations have more room to breath down there.

tck/README.md
ktoso: good catch, thanks

tck/README.md
ktoso: not intentional, thanks!

tck/README.md
ktoso: Agreed, I'd go with this rewording though: `// returning `null` means that the tests covering a failed publisher will be skipped`

viktorklang: Good suggestion, perhaps even: `// returning null means that the tests validating a failed publisher will be skipped`

ktoso: sounds good to me :)
updating the PR slowly now

tck/README.md
ktoso: Updated a bit. Using we and you here, seems neutral to me in this section.. did I miss something?

ktoso: Ah of course, you mean avoiding "you", right?

ktoso:
Rebased and addressed comments.

tck/README.md
```diff
@@ -32,53 +32,65 @@ The TCK is provided as binary artifact on [Maven Central](http://search.maven.or
...
+Specification rule abides the following naming convention: `TYPE_spec###_DESC` where:
 
+* `TYPE` is one of: [#type-required](required), [#type-optional](optional), [#type-stochastic](stochastic) or [#type-untested](untested) which describe if this test is covering a Rule that MUST or SHOULD be implemented. The specific words are explained in detail below.
 * `###` is the Rule number (`1.xx` Rules are about Publishers, `2.xx` Rules are about Subscribers etc.)
 * `DESC` is a short explanation of what exactly is being tested in this test case, as sometimes one Rule may have multiple test cases in order to cover the entire Rule.
 
+Here is an example test method signature:
+
 ```java
   // Verifies rule: https://github.com/reactive-streams/reactive-streams#1.1
```
viktorklang: These links are dead now with the move to `reactive-streams-jvm`?

tck/README.md
```diff
@@ -32,53 +32,65 @@ The TCK is provided as binary artifact on [Maven Central](http://search.maven.or
...
 ```
 
-The prefixes of the names of the test methods are used in order to signify the character of the test. For example, these are the kinds of prefixes you may find:
-"`required_`", "`optional_`", "`stochastic_`", "`untested_`".
-
-Explanations:
+#### Test types explained:
 
 ```java
 @Test public void required_spec101_subscriptionRequestMustResultInTheCorrectNumberOfProducedElements() throws Throwable
```
viktorklang: I'm slightly concerned with having actual rule test names embedded in the documentation as there's a real risk of staleness, but perhaps this is something we can live with for now.

tck/README.md
ktoso: Actually not "really dead" - github redirects :-) Excellent point though, will change.

viktorklang: Both points are good news to me! :)

tck/README.md
```diff
@@ -32,53 +32,65 @@ The TCK is provided as binary artifact on [Maven Central](http://search.maven.or
...
+Usually this means that this test case can yield false positives ("be green") even if for some case, the given implementation may violate the tested behaviour.
 
 ```java
 @Test public void untested_spec106_mustConsiderSubscriptionCancelledAfterOnErrorOrOnCompleteHasBeenCalled() throws Throwable
 ```
 
-... means that the test case is not implemented, either because it is inherently hard to verify (e.g. Rules which use the wording "*SHOULD consider X as Y*") or have not been implemented yet (though we hope we have implemented all we could!). Such tests will show up in your test runs as `SKIPPED`, with a message pointing out that the TCK is unable to validate this Rule. We would be delighted if you can figure out a way to deterministically test Rules, which have been marked with this prefix – pull requests are very welcome!
+<a name="type-untested"></a>
+The `untested_` means that the test case is not implemented, either because it is inherently hard to verify (e.g. Rules which use
+the wording "*SHOULD consider X as Y*") or have not been implemented yet (though we hope we have implemented all we
```
viktorklang: " or have not been implemented yet (though we hope we have implemented all we
+could!)." This section adds nothing. We've added all non-hard tests, which is covered by the previous sentence.

tck/README.md
```diff
@@ -188,14 +200,17 @@ public class RangePublisherTest extends PublisherVerification<Integer> {
 
 Notable configuration options include:
 
-* `maxElementsFromPublisher` – which should only be overridden in case the Publisher under test is not able to provide arbitrary length streams, e.g. it's wrapping a `Future<T>` and thus can only publish up to 1 element. In such case you should return `1` from this method. It will cause all tests which require more elements in order to validate a certain Rule to be skipped,
-* `boundedDepthOfOnNextAndRequestRecursion` – which should only be overridden in case of synchronous Publishers. This number will be used to validate if a
-`Subscription` actually solves the "unbounded recursion" problem (Rule 3.3).
+* `maxElementsFromPublisher` – which should only be overridden in case the Publisher under test is not able to provide
```
viktorklang: Perhaps change this wording to say that it must be implemented if the Publisher is bounded? (avoid using negative language)

tck/README.md
```diff
@@ -188,14 +200,17 @@ public class RangePublisherTest extends PublisherVerification<Integer> {
...
+  arbitrary length streams, e.g. it's wrapping a `Future<T>` and thus can only publish up to 1 element. In such case you
+  should return `1` from this method. It will cause all tests which require more elements in order to validate a certain
+  Rule to be skipped,
+* `boundedDepthOfOnNextAndRequestRecursion` – which should only be overridden in case of synchronous Publishers.
```
viktorklang: This explanation does not really assist the reader in understanding what to set it to, so either that's described on that method, in which case we can point the reader to that, otherwise I'm afraid we need to elucidate the reader a bit more here.

tck/README.md
```diff
@@ -188,14 +200,17 @@ public class RangePublisherTest extends PublisherVerification<Integer> {
...
+  This number will be used to validate if a `Subscription` actually solves the "unbounded recursion" problem (Rule 3.3).
```
viktorklang: I think it's less about solving a problem and more about adhering to a rule :)

tck/README.md
```diff
@@ -453,10 +467,13 @@ public class MyIdentityProcessorVerificationTest extends IdentityProcessorVerifi
 
   @Override
   public Publisher<Integer> createFailedPublisher() {
+    // return Publisher that just signals onError instead of null to run additional tests
```
viktorklang: s/just/always

viktorklang:
:+1: aside from comments!

@reactive-streams/contributors Please give your verdict on these TCK documentation improvements if inclined to do so. I'd want to merge this in as soon as comments are addressed so we can cut the next RC. Thanks

tck/README.md
ktoso: In general the code samples in docs are problematic here.. let's keep as is for now?

ktoso:
Thanks for applying the review fixups @viktorklang! I rebased so it's ready to merge after validation.

viktorklang:
Thanks @ktoso

@reactive-streams/contributors Merging these TCK documentation improvments

PR #250, opened 2015-03-25
+tck explains createElement in more useful terms

### LINKED ISSUE #231: Clarification on SubscriberBlackboxVerification.createElement
Hello folks,

I'm looking at getting the TCK working against Vert.x.

Currently trying to extend SubscriberBlackboxVerification, and I'm not sure how to implement the createElement() method.

@Override
  public Buffer createElement(int i) {
    return ??
  }

In particular I can't figure out what the parameter i represents (I did read the javadoc but none the wiser)... and how it relates to the return value. (We are using streams of Buffer throughout the TCK tests - Buffer is a Vert.x class)

Any pointers?

ktoso:
resolves #231

### DIFF
--- a/tck/README.md
+++ b/tck/README.md
@@ -241,18 +241,29 @@ Note that explicitly passed in values take precedence over values provided by th
 
 Subscriber rules Verification is split up into two files (styles) of tests.
 
-The Blackbox Verification tests do not require the implementation under test to be modified at all, yet they are *not* able to verify most rules. In Whitebox Verification, more control over `request()` calls etc. is required in order to validate rules more precisely.
+The Blackbox Verification tests do not require the implementation under test to be modified at all, yet they are *not* able
+to verify most rules. In Whitebox Verification, more control over `request()` calls etc. is required in order to validate rules more precisely.
+
+It is highly recommended to implement the `SubscriberWhiteboxVerification<T>` instead of the blackbox one even if it is
+more work to do so, as it can test far more corner cases in implementations that would otherwise be left untested
+(if only blackbox tests would be used).
 
 ### createElement and Helper Publisher implementations
 Since testing a `Subscriber` is not possible without a corresponding `Publisher` the TCK Subscriber Verifications
 both provide a default "*helper publisher*" to drive its test and also alow to replace this Publisher with a custom implementation.
 The helper publisher is an asynchronous publisher by default - meaning that your subscriber can not blindly assume single threaded execution.
 
+When extending subscriber verification classes a type parameter representing the element type passed through the stream must be given.
+Implementations usually should not be sensitive to the type of element being signalled, but sometimes a Subscriber may be limited to only be able to work within a known set of types -
+like a `FileSubscriber extends Subscriber<String>` for example, that writes each element it receives as new line into a file.
+For element type agnostic Subscribers it is the simplest to parameterize the tests using `Integer` and in the `createElement(int idx)` method (explained below in futher detail), return the incoming `int`. 
+In case an implementation needs to work on a specific type, the verification class should be parameterized using that type (e.g. `class StringSubTest extends SubscriberWhiteboxVerification<String>`) and a `String` must be returned from the `createElement` method.
+
 While the `Publisher` implementation is provided, creating the signal elements is not – this is because a given Subscriber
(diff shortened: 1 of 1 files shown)


tck/README.md
```diff
@@ -243,11 +243,17 @@ Subscriber rules Verification is split up into two files (styles) of tests.
 
 The Blackbox Verification tests do not require the implementation under test to be modified at all, yet they are *not* able to verify most rules. In Whitebox Verification, more control over `request()` calls etc. is required in order to validate rules more precisely.
 
+It is highly recommended to implement the `SubscriberBlackboxVerification<T>` instead of the blackbox one even if it is more troublesome to do so, as it can test far more corner cases in implementations that would otherwise be left untested (if only blackbox tests would be used).
```
viktorklang: Seems odd to recommend the blackbox instead of... the blackbox...

viktorklang: I suspect the correct word is "work" rather than "troublesome".

tck/README.md
```diff
@@ -243,11 +243,17 @@ Subscriber rules Verification is split up into two files (styles) of tests.
...
+
 ### createElement and Helper Publisher implementations
 Since testing a `Subscriber` is not possible without a corresponding `Publisher` the TCK Subscriber Verifications
 both provide a default "*helper publisher*" to drive its test and also alow to replace this Publisher with a custom implementation.
 The helper publisher is an asynchronous publisher by default - meaning that your subscriber can not blindly assume single threaded execution.
 
+When extending subscriber verification classes a type parameter representing the element type passed through the stream must be given. MostImplementations usually should not be sensitive to the type of element being signalled, but sometimes a Subscriber may be limited to only be able to work within a known set of types - like a `FileSubscriber extends Subscriber<String>` for example, that writes each element it receives as new line into a file.
```
viktorklang: MostImplementations?

tck/README.md
```diff
@@ -243,11 +243,17 @@ Subscriber rules Verification is split up into two files (styles) of tests.
...
 The helper publisher is an asynchronous publisher by default - meaning that your subscriber can not blindly assume single threaded execution.
 
+When extending subscriber verification classes a type parameter representing the element type passed through the stream must be given. MostImplementations usually should not be sensitive to the type of element being signalled, but sometimes a Subscriber may be limited to only be able to work within a known set of types - like a `FileSubscriber extends Subscriber<String>` for example, that writes each element it receives as new line into a file. 
+For element type agnostic Subscribers it is the simplest to parameterize the tests using `Integer` and in the `createElement(int idx)` method (explained below in futher detail), return the incoming `int`. 
+In case an implementation needs to work on a specific type, the verification class should be parameterized using that type (e.g. `class StringSubTest extends SubscriberBlackboxVerification<String>`) and a `String` must be returned from the `createElement` method.
```
viktorklang: Perhaps avoid using blackbox verification in the examples here if we want to encourage the use of whitebox?

tck/README.md
```diff
@@ -268,8 +274,8 @@ public class MySubscriberTest extends SubscriberBlackboxVerification<Integer> {
 ```
 
 
-The `createElement` method MAY be called from multiple
-threads, so in case of more complicated implementations, please be aware of this fact.
+The `createElement` method MAY be called from multiple threads, so in case of more complicated implementations,
```
viktorklang: method MAY be called **concurrently** from multiple threads, so in case of more complicated implementations,

tck/README.md
ktoso: 1) thanks!
2) "if it is more troublesome to do so" sounds valid to me, but I'll reword/simplify :)

tck/README.md
ktoso: +1

tck/README.md
ktoso: important word indeed, thanks!

ktoso:
Thanks for the review, pushed fixups. // @viktorklang

ktoso:
// Note to self, editing anything that needs attention in the github edit+pr window is a bad idea ;)

viktorklang:
LGTM!

@reactive-streams/contributors This is a TCK documentation update addressing a question from @purplefox, merging this in 24h unless someone complains (so we can get RC4 out tomorrow).

viktorklang:
@reactive-streams/contributors Merging without objection

PR #255, opened 2015-03-30
Travis PR validation using both JDK 6 and 8

### LINKED ISSUE #254: Enable JDK8 PR validation to avoid breaking build for Java8
Since Java 8 is more strict about JavaDocs projects which compile under 6 / 7, may not necessarily compile under J8, causing unnecessary trouble.

We should validate the build on both JDK 6 (to avoid accidental use of JDK8 features) and JDK 8 (to make sure it's compile-able under 8).

Most recently problem was bumped into during: https://github.com/reactive-streams/reactive-streams-jvm/pull/252#issuecomment-87188435

> (Aside: One small issue, not specifically related to this PR, I had to install Java 7 to build the TCK as it doesn't build using Java 8 - small JavaDoc issue. Java 7 will be EOL in a month)

ktoso:
By validating on both JDKs we know the project even builds on 8,
while not using features (classes) from JDK8 - so it's still usable for JDK6 projects.

Resolves #254

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ cache:
   directories:
   - $HOME/.gradle
 jdk:
+  - oraclejdk8
   - openjdk6
 env:
   global:


viktorklang:
Makes total sense, thanks @ktoso.

@reactive-streams/contributors Merging this as is as I have a hard time coming up with any counter arguments for making sure that RS is usable with both Java 6 and 8.

Issue #259, opened 2015-04-04
PublisherVerification#required_spec317_mustNotSignalOnErrorWhenPendingAboveLongMaxValue takes a long time for fully synchronous publishers


alkemist:
Hi,

I can't quite determine whether this is a defect in my test, or my interpretation of the spec. 

I have an Iterable publisher, that is fully synchronous. With this test, it effectively tries to publish 0..Integer.MAX_VALUE which takes some time. 

This won't return until all of the values have been published: 

https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java#L1057

Solutions I can see:

1. Use something more reasonable than Integer.MAX_VALUE.
2. Cancel the subscription when `callsCounter == 0`

Or of course, my implementation is wrong.

smaldini:
I think the intent is more about testing the pending request handling which can wrap if you request more long than long max (negative number) , which is going to fail somehow since negative demand is prohibited. The publisher helper should be passed an unlimited marker, at this stage you should not necessarily map the request to number of elements if it's Long.Max for the purpose of the test.

ktoso:
Hi guys,
@alkemist you are right that a naive synchronous implementation will keep running forever here. 
I think we can indeed improve this test for purely synchronous Publishers via your 2nd proposal, but I'd like to draw out the bigger picture about this test and how we handle it in Akka as well:

The purpose of this test is indeed to test the overflowing of demand (as @smaldini points out) - in order to check the Publisher doesn't `onError` or fail otherwise if demand overflows [3.17 - last sentence](https://github.com/reactive-streams/reactive-streams-jvm#3.17). We could use smaller numbers in each request and request more times... I don't feel like it's that is changing much though. 

@alkemist I do agree that we can send the `cancel()` inside [here](https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java#L1047) and this might help some synchronous Publisher implementations (those that during their "loop sending onNext signals" check for cancellation). Do you want to send in the PR?

[...]

PR #260, opened 2015-04-05
Cancel the subscription after receiving all of the pertinent emissions (#259)

alkemist:
See #259.

### DIFF
--- a/tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
+++ b/tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
@@ -5,6 +5,7 @@ import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
 import org.reactivestreams.tck.support.TCKVerificationSupport;
 import org.reactivestreams.tck.support.TestException;
+import org.testng.Assert;
 import org.testng.annotations.Test;
 
 import java.util.Random;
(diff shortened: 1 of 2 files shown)


alkemist:
@ktoso can you provide some guidance on how you want this tested please.

ktoso:
I think adding a test that has a simple synchronous publisher that checks for the cancellation signal while spinning would be good enough here - we just want to show that such test will properly terminate - right?

ktoso:
Got a few mins more time, more hints which should help you get started :-)

Put the test inside `PublisherVerificationTest`, the used naming format is `testedMethodName_shouldPass[_optionalDesctiptionWhen]`:

In our case I'd go with:

```scala
@Test
  public void required_spec317_mustNotSignalOnErrorWhenPendingAboveLongMaxValue_forSynchronousPublisher() throws Throwable {
```

viktorklang:
:+1:

alkemist:
@ktoso ping.

tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
```diff
@@ -585,6 +586,38 @@ public void required_spec317_mustSupportACumulativePendingElementCountUpToLongMa
...
+          @Override
+          public void cancel() {
+            cancelled = true;
+          }
+        });
+      }
+    }).required_spec317_mustNotSignalOnErrorWhenPendingAboveLongMaxValue();
+
+    // 11 due to the implementation of this particular TCK test (see impl)
+    Assert.assertEquals(sent.get(), 11);
```
ktoso: I think I'd skip this check, the test "passes" if it doesn't end up infinitely publishing - it's true that it is 11 signals but I wouldn't test for this number.

ktoso:
Yup good one, LGTM (I think I'd skip the `11` check, it's a bit internals looking, no need for it IMO). Thanks @alkemist

tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
viktorklang: Since it's the verification test, isn't it fine to be specific?

ktoso: No strong feelings about it, can stay as is :)

viktorklang: If it is _wrong_ I'd want to see it removed, but from a cursory inspection I don't see that it could be problematic, but if I am wrong, please do tell me!

ktoso: It's definitely not wrong, just depending on the internally+arbitrarily chosen value inside of `required_spec317_mustNotSignalOnErrorWhenPendingAboveLongMaxValue`.
You're right though that these are very specific tests so keeping as-is isn't a bad idea.

viktorklang: If we change the underlying TCK impl then we'd also have to change this assertion, I think that's good.

alkemist:
I've already done the copyright waiver thing: https://github.com/reactive-streams/reactive-streams-jvm/blob/master/CopyrightWaivers.txt#L29

viktorklang:
@reactive-streams/contributors Let's merge this TCK fix in and issue a new RC immediately to get feedback.

viktorklang:
Merging

PR #262, opened 2015-04-08
-tck rm undocumented and unused publisherReferenceGCTimeoutMillis method

ktoso:
Removes unused and not documented method which may have lead to confusion on how to set up GC-timeout.

GC timeout it set up during creation of a verification, the same style as the default timeout in TestEnvironment, as documented here: https://github.com/reactive-streams/reactive-streams-jvm/commit/b17a364911cbde7bb594a490dd6159b0c8ada20a#diff-4ce068920e11452416705ccb8f46ec78R198

This PR removes an unused, configuring this value should be done as documented in tck/README.md


viktorklang:
:+1:

@reactive-streams/contributors Removes unused and unneeded method from the TCK, targeting RC5 for this.

viktorklang:
Merging

Issue #223, opened 2015-02-17
Release 1.0.0


viktorklang:
@reactive-streams/contributors 

Preparations for the release:

- [x] Update http://www.reactive-streams.org/ to only list libraries/implementations that pass the final TCK & spec

- [x] Add the list of libraries/implementations that pass the final TCK & spec to the README.md

- [X] Do a final pass at proofreading & spellchecking the documentation both for the TCK and the javadoc etc.

- [X] Update all references to the Reactive Streams version in `https://github.com/reactive-streams/reactive-streams-jvm` to reflect `1.0.0`

- [X] Run the build

- [X] Run the tests

- [X] Create the jars (code, source, javadoc)

- [X] Publish the jars to Maven Central

- [x] Draft announcement

Anything else?

viktorklang:
@reactive-streams/contributors FYI, I'll be on vacation the coming 2+ weeks (until the 9th of March) and so for things that cannot wait until then, I'll have to delegate to the inimitable @rkuhn.

rkuhn:
I’d vote for #224 being decided upon before we do 1.0.0.final.

viktorklang:
@reactive-streams/contributors with 1.0.0.RC5 being released minutes ago, it is time to try it out and verify that it works for your implementations. We want to ship 1.0.0.final STAT, ending the 4+ months RC cycle with amazing feedback and improvements in both the spec, TCK and documentation.

viktorklang:
@reactive-streams/contributors Please report back here after updating to RC5!

ktoso:
Akka successfully updated to RC5: https://github.com/akka/akka/pull/17166

alkemist:
Ratpack and Rx are now building against with rc5.

viktorklang:
@alkemist Thanks for the update! Now we're just waiting for @smaldini & @purplefox :)

viktorklang:
@reactive-streams/contributors I think we ought to have a demo of the different implementations interoperating to show when we cut 1.0.0.final. Who volunteers to make this happen?

alkemist:
I recall there being a RxJava, Akka and Ratpack demo somewhere. I think it might have been done by @rkuhn or someone else on the Akka team for a presentation.

ktoso:
It's https://github.com/rkuhn/ReactiveStreamsInterop Just needs dependency updating and maybe a line or two of changes I guess... :-) // cc @rkuhn

[...]

PR #267, opened 2015-04-24
update ref to 1.0.0.final

smaldini:
WIP for #223

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,12 +8,12 @@ The latest preview release is available on Maven Central as
 <dependency>
   <groupId>org.reactivestreams</groupId>
   <artifactId>reactive-streams</artifactId>
-  <version>1.0.0.RC5</version>
+  <version>1.0.0.final</version>
 </dependency>
 <dependency>
   <groupId>org.reactivestreams</groupId>
   <artifactId>reactive-streams-tck</artifactId>
-  <version>1.0.0.RC5</version>
+  <version>1.0.0.final</version>
   <scope>test</scope>
 </dependency>
 ```
--- a/README.md
+++ b/README.md
@@ -8,12 +8,12 @@ The latest preview release is available on Maven Central as
 <dependency>
   <groupId>org.reactivestreams</groupId>
   <artifactId>reactive-streams</artifactId>
-  <version>1.0.0.final</version>
+  <version>1.0.0</version>
 </dependency>
 <dependency>
   <groupId>org.reactivestreams</groupId>
   <artifactId>reactive-streams-tck</artifactId>
-  <version>1.0.0.final</version>
+  <version>1.0.0</version>
   <scope>test</scope>
 </dependency>
 ```
--- a/build.gradle
+++ b/build.gradle
@@ -39,10 +39,10 @@ subprojects {
 
     jar {
         manifest {
-            instruction "Bundle-Vendor", "Reactive Streams SIG"
-            instruction "Bundle-Description", "Reactive Streams API"
-            instruction "Bundle-DocURL", "http://reactive-streams.org"
-            instruction "Bundle-Version", "1.0.0"
+            instructionReplace "Bundle-Vendor", "Reactive Streams SIG"
+            instructionReplace "Bundle-Description", "Reactive Streams API"
+            instructionReplace "Bundle-DocURL", "http://reactive-streams.org"
+            instructionReplace "Bundle-Version", "1.0.0"
         }
     }
 
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ subprojects {
             instructionReplace "Bundle-Vendor", "Reactive Streams SIG"
             instructionReplace "Bundle-Description", "Reactive Streams API"
             instructionReplace "Bundle-DocURL", "http://reactive-streams.org"
-            instructionReplace "Bundle-Version", "1.0.0"
+            instructionReplace "Bundle-Version", "1.0.0.release"
         }
     }
 
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ subprojects {
     apply plugin: "osgi"
 
     group = "org.reactivestreams"
-    version = "1.0.0.RC5"
+    version = "1.0.0.final"
 
     sourceCompatibility = 1.6
     targetCompatibility = 1.6
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ subprojects {
     apply plugin: "osgi"
 
     group = "org.reactivestreams"
-    version = "1.0.0.final"
+    version = "1.0.0"
 
     sourceCompatibility = 1.6
     targetCompatibility = 1.6
@@ -36,11 +36,13 @@ subprojects {
         mavenCentral()
     }
 
+
     jar {
         manifest {
             instruction "Bundle-Vendor", "Reactive Streams SIG"
             instruction "Bundle-Description", "Reactive Streams API"
             instruction "Bundle-DocURL", "http://reactive-streams.org"
+            instruction "Bundle-Version", "1.0.0"
         }
     }
 
--- a/tck/README.md
+++ b/tck/README.md
@@ -27,7 +27,7 @@ The TCK is provided as binary artifact on [Maven Central](http://search.maven.or
 <dependency>
   <groupId>org.reactivestreams</groupId>
   <artifactId>reactive-streams-tck</artifactId>
-  <version>1.0.0.RC5</version>
+  <version>1.0.0.final</version>
   <scope>test</scope>
 </dependency>
 ```
--- a/tck/README.md
+++ b/tck/README.md
@@ -27,7 +27,7 @@ The TCK is provided as binary artifact on [Maven Central](http://search.maven.or
 <dependency>
   <groupId>org.reactivestreams</groupId>
   <artifactId>reactive-streams-tck</artifactId>
-  <version>1.0.0.final</version>
+  <version>1.0.0</version>
   <scope>test</scope>
 </dependency>
 ```


viktorklang:
Looks good so far :)

smaldini:
Any feedback on the grammar, I think so far so good on this side but I feel a bit dodgy on this aspect to check the related box :)

viktorklang:
@smaldini I've done a pass, but I'd like to have more eyes on it.

smaldini:
maybe we can start the release of the binary and checks the impls too.

smaldini:
Right let's start the testing process, should I publish final on codehaus before merging @viktorklang ?

viktorklang:
@smaldini I'm not sure, assuming we already validated the integrity of the bits (perhaps do a RC5-final binary comparison before pushing?) it seems like we should be able to push it to maven central.
Wdyt?

smaldini:
I'm usually risk averse and I would just merge and push to maven.

viktorklang:
@smaldini Check the SHA of the RC5 jars and compare to the SHA of the final jars before pushing, and I'll sleep better :)

smaldini:
I've rarely seen 1.0.0 that works for everyone anyway, thats why there is a minor digit hey :dancer: 
But I will test SHA

smaldini:
there's a bunch of TCK javadoc missing but well...

viktorklang:
@smaldini documentation improvement is never done :)

smaldini:
Strange , .asc and md5 are different from version to another.

smaldini:
which is strange since I'm only one commit away. looking at the content

smaldini:
right Manifest is regenerated

viktorklang:
@smaldini Alright, if that's the only diff then let's go :)

smaldini:
It is on its way to maven central, joyful tear.

viktorklang:
Awesome, @reactive-streams/contributors should update to it so we can
announce shortly.

-- 
Cheers,
√

smaldini:
http://repo1.maven.org/maven2/org/reactivestreams/reactive-streams/1.0.0.final/

viktorklang:
@jsuereth and @rkuhn suspect that the ".final" part of the version number
may trip up version conflict resolution.
Since the ".final" serves no purpose, I propose we pull the "1.0.0.final"
and republish it as "1.0.0" while we still can. Is that OK, @smaldini?

-- 
Cheers,
√

alkemist:
+1

rkuhn:
@alkemist The problem would be that OSGi considers "1.0.0" < "1.0.0.RC5", how do we set a different version in the OSGi manifest, something like "1.0.0.final" is actually the right thing for OSGi.

alkemist:
Still +1.

rkuhn:
background on why using "1.0.0.final" for Maven is problematic: http://mojo.codehaus.org/versions-maven-plugin/version-rules.html (in short: "1.0.0.final" is interpreted as "0.0.0-1.0.0.final" and effectively only string comparison is used)

rkuhn:
@alkemist I take it that your +1 is for the "1.0.0" maven version. In sbt I would know how to fix the OSGi version, how do we do this in gradle?

alkemist:
FWIW, Ivy (therefore SBT) and Gradle understand that final > rc.

benjchristensen:
I also think we should release 1.0.0 and then if OSGi is a mess release 1.0.1 as a patch release.

alkemist:
@rkuhn what exactly do you want to do? Just use `1.0.0.final` in the OSGi manifest?

rkuhn:
Yes, that would solve all the issues AFAICS.

viktorklang:
Solving all issues sounds good to me

-- 
Cheers,
√

viktorklang:
So, what's the action plan:

What specifically needs to be done?
How do we verify that it works?
Who does what?
When?

Could someone please confirm that the 1.0.0.final bits have been pulled,
STAT? @smaldini or @rkhun should have access

-- 
Cheers,
√

viktorklang:
@rkuhn that is

-- 
Cheers,
√

viktorklang:
>FWIW, Ivy (therefore SBT) and Gradle understand that final > rc.

@alkemist Does it understand that "final" > "RC<n>"?

viktorklang:
I noticed that the bits are already in the release repo and not in the staging repo. I don't think we'll be able to pull them at this point. We could contact Sonatype and investigate whether they would be able to help us out. Thoughts?

rkuhn:
To me the situation presents itself like this:

* we must release a version that compares greater to the currently highest Maven version, which surprisingly is `0.4.0` at this point (since all our other releases are `0.0.0-<qualifier>`)
* we must release a version that compares greater to the currently highest OSGi version, which is `1.0.0.final` at this point

The simples solution would be `1.0.1`, but if we don’t like that then we can release `1.0.0` for Maven with an OSGi manifest that says `1.0.0.release`. Then we don’t need to pull artifacts and we still don’t get version collisions.

Now to the actions: I can easily execute the `1.0.1` release. If we want to go the `1.0.0` route then someone else will have to provide a patch for the gradle build to make it happen.

viktorklang:
@rkuhn

Releasing `1.0.0` with a `1.0.0.release` manifest would not address the possible confusion between "1.0.0.final" and "1.0.0" for users though. Perhaps we needn't worry if we never advertise `1.0.0.final` or if we add documentation to only use `x.y.z` artifacts.

As for Gradle instructions I think its but @alkemist should confirm:

~~~groovy
 jar {
        manifest {
            instruction "Bundle-Vendor", "Reactive Streams SIG"
            instruction "Bundle-Description", "Reactive Streams API"
            instruction "Bundle-DocURL", "http://reactive-streams.org"
            instruction "Bundle-Version", "1.0.0.release"
        }
    }
~~~

Current OSGi config: https://github.com/reactive-streams/reactive-streams-jvm/blob/master/build.gradle#L43
Docs: http://gradle.org/docs/current/userguide/osgi_plugin.html

rkuhn:
@viktorklang I think that the best option for `1.0.0.final` now is to never mention it anywhere. Then there should be no issues that people run into in practice.

viktorklang:
@rkuhn The version that shall not be named.

In that case, I vote for `1.0.0` with a rectified OSGi version as you proposed.

alkemist:
@viktorklang that Gradle config should work. I'd eyeball the JAR though.

viktorklang:
@alkemist +1 for eyeballing the jar, and also, thanks for verifying my Gradle claims :)

viktorklang:
So, proposed action plan:

* delete the `1.0.0.final` tag in the repo
* add the fix for the OSGi version
* update the other version references to reflect `1.0.0`
* cut the bits
* inspect generated jars for correct OSGi Manifest & version
* push 

Does that sound like a plan? Any objections or suggestions?

smaldini:
Gradle OSGI plugin was already taking care of the bundle version but yeah making explicit is not harmful.

rkuhn:
I tried it out with Viktor’s instructions locally and this was the resulting manifest:
~~~
rk:libs rkuhn$ cat META-INF/MANIFEST.MF 
Manifest-Version: 1.0
Bundle-Description: Reactive Streams API
Export-Package: org.reactivestreams;version="1.0.0",1.0.0.release
Bundle-SymbolicName: org.reactivestreams.reactive-streams
Bundle-Version: 1.0.0,1.0.0.release
Bundle-Name: reactive-streams
Bundle-ManifestVersion: 2
Bnd-LastModified: 1424093440000
Bundle-Vendor: Reactive Streams SIG
Bundle-DocURL: http://reactive-streams.org
Created-By: 1.8.0_31 (Oracle Corporation)
Tool: Bnd-2.1.0.20130426-122213
~~~
I tried to find information about what that version string means but that syntax is not mentioned anywhere in the freely available docs (I do not have an OSGi specification handy).

smaldini:
Yeah Grade OSGI plugin uses the version you specify :)

alkemist:
Ah, use `instructionReplace` instead of `instruction`.

smaldini:
we don't need that extra line

smaldini:
Is that correct enough for another try :) ?
```
Archive:  reactive-streams-1.0.0.jar
   creating: META-INF/
  inflating: META-INF/MANIFEST.MF    
   creating: org/
   creating: org/reactivestreams/
  inflating: org/reactivestreams/Processor.class  
  inflating: org/reactivestreams/Publisher.class  
  inflating: org/reactivestreams/Subscriber.class  
  inflating: org/reactivestreams/Subscription.class  
wifi-5-214:libs smaldini$ cat META-INF/MANIFEST.MF 
Manifest-Version: 1.0
Bundle-Description: Reactive Streams API
Export-Package: org.reactivestreams;version="1.0.0"
Bundle-SymbolicName: org.reactivestreams.reactive-streams
Bundle-Version: 1.0.0
Bundle-Name: reactive-streams
Bundle-ManifestVersion: 2
Bnd-LastModified: 1430236558000
Bundle-Vendor: Reactive Streams SIG
Bundle-DocURL: http://reactive-streams.org
Created-By: 1.8.0_20 (Oracle Corporation)
Tool: Bnd-2.1.0.20130426-122213
```

viktorklang:
@smaldini no, "Bundle-Version: 1.0.0" is wrong according to for OSGi we need something like "Bundle-Version: 1.0.0.release" (@rkuhn can confirm)

smaldini:
Who uses OSGI again ? sorry fixing that.

smaldini:
```
  inflating: META-INF/MANIFEST.MF    
  inflating: org/reactivestreams/Processor.class  
  inflating: org/reactivestreams/Publisher.class  
  inflating: org/reactivestreams/Subscriber.class  
  inflating: org/reactivestreams/Subscription.class  

wifi-5-214:libs smaldini$ cat META-INF/MANIFEST.MF 

Manifest-Version: 1.0
Bundle-Description: Reactive Streams API
Export-Package: org.reactivestreams;version="1.0.0.release"
Bundle-SymbolicName: org.reactivestreams.reactive-streams
Bundle-Version: 1.0.0.release
Bundle-Name: reactive-streams
Bundle-ManifestVersion: 2
Bnd-LastModified: 1430236558000
Bundle-Vendor: Reactive Streams SIG
Bundle-DocURL: http://reactive-streams.org
Created-By: 1.8.0_20 (Oracle Corporation)
Tool: Bnd-2.1.0.20130426-122213
```

smaldini:
now ?

viktorklang:
I'm not sure about: "Export-Package: org.reactivestreams;version="1.0.0.release""
@rkuhn?

For the record, let it be known, OSGi is an instrument for inflicting pain.

smaldini:
I'm going to make a slide about at springio.net tomorrow I think...

viktorklang:
@smaldini a slide about what? OSGi?

smaldini:
Yeah :+1: who cares about specs that are actually changing something without involving operating systems type containers.

rkuhn:
@smaldini Your latest manifest is looking good!

viktorklang:
@rkhun Just to make sure, is "1.0.0.release" considered to be newer than "1.0.0.RC5" by OSGi? (note difference in upper/lower case)

rkuhn:
Yes, it is, see https://osgi.org/javadoc/r4v43/core/org/osgi/framework/Version.html (it uses `String.compareTo`).

viktorklang:
@rkuhn Thanks!

Alright, so it looks like we have a solution!

let's switch to version `1.0.0` with OSGi version being `1.0.0.release` in the repo
generate the jars
inspect the jars to make sure they are OK
publish `1.0.0`
delete the `1.0.0.final` tag from the `reactive-streams-jvm` repository

Sounds good?

smaldini:
The tag is not here yet since the PR was not merged. Right now it can then :)

viktorklang:
@rkuhn @smaldini @benjchristensen @alkemist and @reactive-streams/contributors in general.

I think we ought to merge this PR and publish the fixed `1.0.0` ASAP so we can get unstuck in the release process. @rkuhn, since you discovered the problem and found an appropriate fix, would you like to do the honors of cutting the release, or should our "release general" @smaldini do it?

rkuhn:
Yes, I’ll start with the release now.

viktorklang:
Godspeed

-- 
Cheers,
√

smaldini:
Thanks @rkuhn! I was literally in flight from London Publisher to Barcelona Subscriber, fortunately flights are using backpressure too.

viktorklang:
@smaldini There is back pressure for sure, the longer the flight the worse back pain...

PR #275, opened 2015-06-20
add CC0 label to README

rkuhn:
To make it visually obvious that CC0 applies to this work we might want to add this standard marker. What do you think, @reactive-streams/contributors ?


benjchristensen:
:+1:

viktorklang:
:+1:

PR #289, opened 2015-08-20
Fix 'foreach' comment in AsyncSubscriber and rename SyncSybscriber.foreach to whenNext

rstoyanchev:
AsyncSubscriber had a comment referring to `foreach` which is a method that exists in SyncSubscriber while in AsyncSubscriber it is called `whenNext`.

This PR also renames `foreach` in SyncSubscriber to `whenNext`. When I first looked at SyncSubscriber I had the impression there was a loop. It doesn't make sense I know since I did look at `foreach`. Nevertheless it was only when I debugged that I realized that just above is `if (!done)` and not `while (!done)`.

Feel free to ignore this if you had some strong reason for the naming.


viktorklang:
@rstoyanchev Thanks for the correction/touch-up! Would you mind adding yourself to the copyright-waivers in this commit and push again so that it can be merged? Thanks, √

rstoyanchev:
Sure thing, done.

ktoso:
LGTM from my point of view as well, thanks :)

viktorklang:
:+1:

viktorklang:
Thank you @rstoyanchev!

PR #296, opened 2015-11-21
Bridge between Reactive-Streams and JDK 9 Flow API

akarnokd:
This PR adds the code to bridge the Reactive-Streams API and the JDK 9 Flow APIs.

Unfortunately, I don't have any idea how to enable JDK 9 compilation or if it is even possible with gradle/travis at this point.


alkemist:
Gradle has limited support for it. Compilation works, but executing tests with 9 does not. It's in progress, no ETA.

viktorklang:
Thanks @akarnokd! Let's keep this here until we find a way to make Gradle not stand in our way :)

akarnokd:
Project Reactor uses a JSR166 backport for `Flow` interactions; I've added a dependency for it. Build for Java 8 passes but the backport won't work with Java 6.

viktorklang:
@akarnokd thanks for the update! I guess Gradle and Java9 support is still not on the horizon?

viktorklang:
@alkemist Do you happen to know if Gradle works with java 9 now?

akarnokd:
@viktorklang Gradle 3+ should work, but that's 1/3 of the problem: another problem is that Travis doesn't support Java 9 builds yet afaik plus this project's setup requires a Java 6 compatible source in every subproject and currently it fails with `class file has wrong version 52.0, should be 50.0`.

viktorklang:
@akarnokd If we cannot update the build file to support jdk9 we could most definitely create a subproject and have that properly configured. About TravisCI there's: https://github.com/travis-ci/travis-ci/issues/5520#issuecomment-175914595

akarnokd:
I've updated the PR to include conditional building of the Java 9 parts. Unfortunately, Travis-CI was not too cooperative; sorry about the commit spams. A few highlights:

- Java 9 build requires the `trusty` CI environment
- Java 9 requires Gradle 4+
- `trusty` doesn't support `openjdk6`
- Jdk 6 and 7 can't download Gradle 4+ through https due to cipher issues
- couldn't figure out if there is a way to run `openjdk6` on `normal` and the rest on `trusty` via the yml
- Gradle 4+ doesn't support jdk 6

.travis.yml
```diff
@@ -1,13 +1,35 @@
...
+        - tar -xzf jdk-9+177_linux-x64_bin.tar.gz
+        - export JAVA_HOME=~/jdk-9
+        - PATH=$JAVA_HOME/bin:$PATH
+        - cd -
+
+# Don't let Travis CI execute './gradlew assemble' by default
+install:
+# Display Gradle, JVM and other versions
+  - ./gradlew -version
+
```
ktoso: Copied from my PR, could attribute as commit?
https://github.com/reactive-streams/reactive-streams-jvm/pull/383

build.gradle
```diff
@@ -3,7 +3,7 @@ subprojects {
     apply plugin: "osgi"
 
     group = "org.reactivestreams"
-    version = "1.0.1-RC2"
+    version = "1.0.0"
```
ktoso: why?

ktoso: For back releasing we'd do so separately if needed

settings.gradle
```diff
@@ -1,9 +1,27 @@
...
+    println("Java 9 Flow API not available")
+}
+
 include ':reactive-streams'
 include ':reactive-streams-tck'
 include ':reactive-streams-examples'
 
+if (jdkFlow) {
+    include ':reactive-streams-flow-bridge'
+}
```
ktoso: ok, nice 👍

flow-bridge/src/main/java/org/reactivestreams/ReactiveFlowBridge.java
```diff
@@ -0,0 +1,357 @@
...
+ * You should have received a copy of the CC0 legalcode along with this  *
+ * work. If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.*
+ ************************************************************************/
+
+package org.reactivestreams;
+
+import java.util.concurrent.Flow;
+
+/**
+ * Bridge between Reactive-Streams API and the Java 9 Flow API.
```
ktoso: javadoc link?

flow-bridge/src/main/java/org/reactivestreams/ReactiveFlowBridge.java
```diff
@@ -0,0 +1,357 @@
...
+ */
+public final class ReactiveFlowBridge {
+    /** Utility class. */
+    private ReactiveFlowBridge() {
+        throw new IllegalStateException("No instances!");
+    }
+
+    /**
+     * Converts a Flow Publisher into a Reactive Publisher.
+     * @param <T> the value type
```
ktoso: `element` to use same wording as elsewhere

flow-bridge/src/test/java/org/reactivestreams/SubmissionPublisherTckTest.java
```diff
@@ -0,0 +1,59 @@
...
+                }
+                sp.close();
+            }
+        }).start();
+        return ReactiveFlowBridge.toReactive(sp);
+    }
+
+    @Override
+    public Publisher<Integer> createFailedPublisher() {
+        return null;
```
ktoso: should be possible to provide one

flow-bridge/src/test/java/org/reactivestreams/SubmissionPublisherTckTest.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public long maxElementsFromPublisher() {
+        return 100;
+    }
+
+}
```
ktoso: Thought: Should we add test classes that work with the JDK9 interfaces directly right away when JDK9 is out?

ktoso:
Naming bikeshed, but important one I think: I would not want to refer to Reactive Streams as "the Reactive one" or "toReactive", it sounds misleading. Even if the alternative sounds boring and long I think it's more correct and less misleading ("wha!? The Java one is not 'Reactive'?" - of course *we* know that's not why the name / not what it means, but it reads like that and adapters may be used by people completely new to the concept)

.travis.yml
```diff
@@ -1,13 +1,35 @@
...
-jdk:
-  - oraclejdk8
-  - openjdk6
+
+before_install:
+  - export GRADLE_OPTS=-Xmx1024m
+
+matrix:
+  include:
+    - jdk: openjdk7
```
ktoso: I think we may be OK not running against 6, but let's see what others say hm

.travis.yml
akarnokd: Okay.

build.gradle
akarnokd: The original PR predated 1.0.1-RC2. I'll restore the version number.

ktoso: Ah I see, thanks :)

flow-bridge/src/main/java/org/reactivestreams/ReactiveFlowBridge.java
akarnokd: Okay.

flow-bridge/src/test/java/org/reactivestreams/SubmissionPublisherTckTest.java
akarnokd: Okay.

flow-bridge/src/main/java/org/reactivestreams/ReactiveFlowBridge.java
akarnokd: Not sure what links do you expect here.

flow-bridge/src/test/java/org/reactivestreams/SubmissionPublisherTckTest.java
akarnokd: Maybe, but perhaps in its own module.

.travis.yml
akarnokd: Not much option here I think, JDK 6 is less and less supported by Gradle/Travis. As long as the compile target is 6, we should be okay.

ktoso: 👍

flow-bridge/src/test/java/org/reactivestreams/SubmissionPublisherTckTest.java
ktoso: Yeah, could be cleaner that way hm... Ticketified https://github.com/reactive-streams/reactive-streams-jvm/issues/388

.travis.yml
```diff
@@ -1,13 +1,37 @@
...
+    - jdk: oraclejdk8 # JDK 1.8.0_131-b11
+    # From https://github.com/reactive-streams/reactive-streams-jvm/pull/383
+    - jdk: oraclejdk9 # JDK 9-ea+174 provided by Travis will be replaced by 9+177 in "before_install"
+      before_install:
+        - cd ~
+        - wget http://download.java.net/java/jdk9/archive/177/binaries/jdk-9+177_linux-x64_bin.tar.gz
```
akarnokd: FYI, looks like this has been deleted and needs to be updated whenever there is a newer JDK build.

viktorklang: If HTTP only supported redirects! ;)

viktorklang:
@reactive-streams/contributors Time to revive this PR? Looks like Travis should support oraclejdk9 OOTB now. @akarnokd, care to update the PR?

.travis.yml
```diff
@@ -1,13 +1,37 @@
...
+  - export GRADLE_OPTS=-Xmx1024m
+
+matrix:
+  include:
+    - jdk: openjdk7
+    - jdk: oraclejdk8 # JDK 1.8.0_131-b11
+    # From https://github.com/reactive-streams/reactive-streams-jvm/pull/383
+    - jdk: oraclejdk9 # JDK 9-ea+174 provided by Travis will be replaced by 9+177 in "before_install"
+      before_install:
+        - cd ~
```
akarnokd: This is no longer needed on Travis, `jdk: oraclejdk9` switches to 9b181 == GA

akarnokd:
Upgrade done.

.travis.yml
```diff
@@ -16,15 +16,7 @@ matrix:
...
-    # From https://github.com/reactive-streams/reactive-streams-jvm/pull/383
-    - jdk: oraclejdk9 # JDK 9-ea+174 provided by Travis will be replaced by 9+177 in "before_install"
-      before_install:
-        - cd ~
-        - wget http://download.java.net/java/jdk9/archive/177/binaries/jdk-9+177_linux-x64_bin.tar.gz
-        - tar -xzf jdk-9+177_linux-x64_bin.tar.gz
-        - export JAVA_HOME=~/jdk-9
-        - PATH=$JAVA_HOME/bin:$PATH
-        - cd -
+    - jdk: oraclejdk9
```
ktoso: Great! :)

ktoso:
Looking good, I'll review later tonight or tomorrow in more depth.
Late here and gotta do some family time

viktorklang:
LGTM

@reactive-streams/contributors Yay or Nay?

flow-bridge/src/main/java/org/reactivestreams/ReactiveFlowBridge.java
ktoso: I meant `{@link java.util.concurrent.Flow}`

flow-bridge/src/main/java/org/reactivestreams/ReactiveFlowBridge.java
```diff
@@ -0,0 +1,357 @@
...
+ * Bridge between Reactive-Streams API and the Java 9 Flow API.
+ */
+public final class ReactiveFlowBridge {
+    /** Utility class. */
+    private ReactiveFlowBridge() {
+        throw new IllegalStateException("No instances!");
+    }
+
+    /**
+     * Converts a Flow Publisher into a Reactive Publisher.
```
ktoso: Please "Reactive Streams Publisher/Subscriber" everywhere, let's be specific; the putting "Reactive" around imprecisely leads to people misunderstanding it and saying "oh yeah but the Flow one is not Reactive!" ;)

flow-bridge/src/main/java/org/reactivestreams/ReactiveFlowBridge.java
```diff
@@ -0,0 +1,357 @@
...
+ ************************************************************************/
+
+package org.reactivestreams;
+
+import java.util.concurrent.Flow;
+
+/**
+ * Bridge between Reactive Streams API and the Java 9 {@link java.util.concurrent.Flow} API.
+ */
+public final class ReactiveFlowBridge {
```
ktoso: Same as commented previously, `ReactiveStreamsFlowBridge`. As this is directly visible user API

flow-bridge/src/main/java/org/reactivestreams/ReactiveFlowBridge.java
```diff
@@ -0,0 +1,357 @@
...
+        }
+        
+    }
+    
+    /**
+     * Wraps a Flow Processor and forwards methods of the Reactive Streams Processor to it.
+     * @param <T> the input type
+     * @param <U> the output type
+     */
+    static final class ReactiveToFlowProcessor<T, U>
```
ktoso: Those can remain `R...Flow` rather than RS I guess, since they're internal and people should not be seeing them directly

flow-bridge/src/test/java/org/reactivestreams/TestConsumer.java
```diff
@@ -0,0 +1,163 @@
...
+package org.reactivestreams;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Flow;
+import java.util.concurrent.TimeUnit;
+
+class TestConsumer<T> implements Flow.Subscriber<T>, Subscriber<T> {
```
ktoso: `TestEitherSubscriber`?

flow-bridge/src/main/java/org/reactivestreams/ReactiveFlowBridge.java
akarnokd: Sure.

flow-bridge/src/main/java/org/reactivestreams/ReactiveFlowBridge.java
akarnokd: Unless we consider stacktraces.

flow-bridge/src/test/java/org/reactivestreams/TestConsumer.java
akarnokd: Okay.

akarnokd:
Updated based on @ktoso's feedback.

build.gradle
```diff
@@ -42,11 +42,11 @@ subprojects {
             instructionReplace "Bundle-Vendor", "Reactive Streams SIG"
             instructionReplace "Bundle-Description", "Reactive Streams API"
             instructionReplace "Bundle-DocURL", "http://reactive-streams.org"
-            instructionReplace "Bundle-Version", "1.0.1-RC2"
+            instructionReplace "Bundle-Version", "1.0.1"
         }
     }
 
-    if (name in ["reactive-streams", "reactive-streams-tck", "reactive-streams-examples"]) {
+    if (name in ["reactive-streams", "reactive-streams-tck", "reactive-streams-examples", "reactive-streams-flow-bridge"]) {
```
viktorklang: Not to start a bikeshedding exercise, but are we on board with the naming of this artifact?

Options are things like "reactive-streams-flow-compat", "reactive-streams-jdk9-compat", "reactive-streams-jdk9-interop" and the like.

ktoso: Heh yeah did not want to open that bikeshed as I've bikeshedded quite a lot here already.
Only precedent for such a library would be the https://github.com/scala/scala-java8-compat library I think, and the inclusion of the number there perhaps is sub optimal. So out of the ones you proposed I like `reactive-streams-flow-compat` I guess, or `reactive-streams-java-flow-compat`.

It's not super important to me though I guess, so would be interesting to hear or do a quick vote I guess?

ktoso:
Once this is merged I'll provide the TCK that can work on the flow interfaces directly btw, so eagerly awaiting this to be merged :)

build.gradle
viktorklang: How about reactive-streams-jsr266-interop?

viktorklang:
@reactive-streams/contributors Given that this has had ample time for reviews, I'm going to merge this. Shall we back-release a version for this for 1.0.0 and 1.0.1 respectively?

akarnokd:
I guess renaming the logical module name can happen afterwards at any time. I'm fine with the merging otherwise. I'd target 1.0.1 only.

PR #306, opened 2016-03-14
Non-BC for TCK: Corrects a typo in test method from *Compuatation to …

### LINKED ISSUE #304: Typo in three places
https://github.com/reactive-streams/reactive-streams-jvm/search?utf8=%E2%9C%93&q=Compuatation

Noticed this when running reactive-kafka tests.

viktorklang:
…*Computation

Fixes #304

@reactive-streams/contributors What do we think, is it OK to do non-BC updates to the TCK for a minor/patch version?

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
@@ -342,8 +342,8 @@ public abstract class IdentityProcessorVerification<T> extends WithHelperPublish
   }
 
   @Override @Test
-  public void untested_spec305_cancelMustNotSynchronouslyPerformHeavyCompuatation() throws Exception {
-    publisherVerification.untested_spec305_cancelMustNotSynchronouslyPerformHeavyCompuatation();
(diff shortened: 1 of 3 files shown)


viktorklang:
@reactive-streams/contributors Ping

viktorklang:
@reactive-streams/contributors Is it fine to do a binary incompatible change in the TCK?

ktoso:
I think we can make such changes. Users are library implementors, and upgrading the TCK may indeed require changing things anyway (e.g. your Publisher now fails etc), so it's not really meant to be a "upgrade and do nothing" upgrade scenario.

I do think however that we should provide a migration guide as we merge such incompatible changes. A simple migration-guide.md would work, that just lists "method name changed from ..." etc. Please include such file stub along with this PR?

The other PR with the thrown Exception type is another one of this kind that we should then add to the markdown file.

viktorklang:
@reactive-streams/contributors Since no one objects to this breaking change (method rename) in the TCK I'll merge this. Sicne the TCK needs to be re-run and this typically happens on tests anyway, bumping the RS version means having to change the  build anyway, s o it may as well also be renaming a couple of methods?

PR #302, opened 2016-02-02
Update to Gradle 2.10

alkemist:

PR #309, opened 2016-03-21
+tck #308 allow configuring "no events during N time" separately

### LINKED ISSUE #308: Separate TCK timeouts for things to happen from things to not happen
The TCK has one timeout that's used for two purpose, it's used as a limit to wait for things to happen, eg assert that a signal was eventually received, and it's used to wait for things to never happen, eg assert that a signal is never received.  The default value is 500ms, and using this timeout the TCK may take, for example, around 20 seconds to run.  Now on most laptops, 500ms is reasonable for both values, the tests don't take too long as a result of waiting for things not to happen, and things almost always happen within 500ms.

However, on CI servers, it's a different story.  Take for example Travis.  Our experience with running Play on Travis shows that timeouts need to be as long as 10 seconds, since travis uses a container architecture that runs many builds on the one server concurrently, and GC pauses can take seconds.  But if you change the timeouts to 10 seconds, then the build can end up taking many minutes, because now it's waiting 10 seconds every time it wants to assert that something doesn't happen.

I think these two timeouts should be separated so they can be individually configured.

ktoso:
Because we all love quicker builds <3

Resolves #308

cc @viktorklang @jroper

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ env:
   global:
   - TERM=dumb
   - DEFAULT_TIMEOUT_MILLIS=300
+  - DEFAULT_NO_SIGNALS_TIMEOUT_MILLIS=200
   - PUBLISHER_REFERENCE_GC_TIMEOUT_MILLIS=300
 addons: # Fix OpenJDK build. Issue: https://github.com/travis-ci/travis-ci/issues/5227 & https://docs.travis-ci.com/user/hostname
   hostname: rs-jvm
--- a/tck/README.md
+++ b/tck/README.md
@@ -209,13 +209,19 @@ within the TCK which await for something to happen. The other timeout is `publis
 [Rule 3.13](https://github.com/reactive-streams/reactive-streams-jvm#3.13) which defines that `Subscriber` references MUST be dropped
 by the Publisher.
 
+Note that the TCK differenciates between timeouts for "waiting for a signal" (``defaultTimeoutMillis``),
+and "asserting no signals happen during a given amount of time" (``envDefaultNoSignalsTimeoutMillis``).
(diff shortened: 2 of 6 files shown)


ktoso:
Build is failing on JDK6 due to same magic as seen in https://github.com/reactive-streams/reactive-streams-jvm/pull/302#issuecomment-197653590
Must be some evil spirits (JVM crashes).

ktoso:
Any chance for help resolving the build problems?
We'd (and others) would very much like this change

viktorklang:
@ktoso Could you rebase this on master?

ktoso:
Yes, rebased


On 5 April 2016 at 17:59:23, Viktor Klang (√) (notifications@github.com) wrote:

@ktoso Could you rebase this on master?

viktorklang:
I'm pushing your branch to the repo and opening a local PR to see if Travis makes it pass.

ktoso:
Hmmm, okey thanks. That'd be uber weird though if it then works hm, let's see

patriknw:
this has not been released yet, right?

viktorklang:
@patriknw No, we're gearing up for a 1.0.1 release but there are a few spec clarifications we need to deal with first. If you have time, since this is TCK-related you could make a local build and verify that the feature works.

drewhk:
Ping @viktorklang

viktorklang:
@drewhk Ah, the 1.0.1 release. We should try to roll that one out. What's the timeline?

drewhk:
No particular timeline for us, just want to reduce build-times and we need this feature from the TCK. I just looked at old tickets and decided to ping.

viktorklang:
SGTM

I am a bit swamped at the moment but hope we can prep the release soonish.

-- 
Cheers,
√

PR #312, opened 2016-04-05
Third attempt at fixing the Travis builds

viktorklang:
One more try… apparently the magic word for sudo is "required", and not "true", whereas to turn it off, it is "false" … :S


viktorklang:
@reactive-streams/contributors Third attempt, now with magical incantation added

PR #314, opened 2016-04-06
+tck #308 allow configuring "no events during N time" separately

viktorklang:
Supersedes #309 because TravisCI is broken

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ env:
   global:
   - TERM=dumb
   - DEFAULT_TIMEOUT_MILLIS=300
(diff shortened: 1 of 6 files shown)


ktoso:
wat!? It actually passes now... only difference being from which branch PR originates? wat?

viktorklang:
@ktoso Yes, there were two Travis Issues, one where it failed in general, and now it only fails for forks. :S

viktorklang:
@reactive-streams/contributors There's been ample time to review and give feedback. Merging this.

ktoso:
> one where it failed in general, and now it only fails for forks. :S

*Whoa...!*

Thanks for handling this one.

PR #332, opened 2016-08-20
adjust Subscription.cancel javadoc

davidmoten:
As I understand it `cancel` can be called synchronously also. Removed the part of the javadoc that says that `cancel` is called asynchronously.

### DIFF
--- a/api/src/main/java/org/reactivestreams/Subscription.java
+++ b/api/src/main/java/org/reactivestreams/Subscription.java
@@ -27,7 +27,7 @@ public interface Subscription {
     /**
      * Request the {@link Publisher} to stop sending data and clean up resources.
      * <p>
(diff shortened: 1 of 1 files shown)


viktorklang:
I think that is a reasonable change!

The question is whether we need to explain why that (the showing up of elements after the invocation to cancel) is the case or not?

viktorklang:
@reactive-streams/contributors wdyt?

ktoso:
I think it's a good change. Anyone implementing these should be aware it could be all async in any case.

DougLea:
Looks good -- the phrase "as this call is asynchronous" was unnecessary and unhelpful.

rkuhn:
LGTM

ktoso:
Ping @viktorklang, seems good to go?

viktorklang:
@ktoso Aye, thanks for the ping!

ktoso:
:+1:

PR #339, opened 2017-01-27
Documentation of intent for each and every rule

viktorklang:
Dear @reactive-streams/contributors,

we've been talking for ages about adding an intents-section for every spec rule,
and during a cross-atlantic flight I finally found the time to write this up.

This PR is not intended for merging, but is rather a conversation to figure out the proper wording/intent for each of the spec rules.

The idea would be to iterate on that, and then I'll create a real PR against the README.md

Additionally, I've found some opportunities to clarify some of the spec rules, and I think that'd be of general interest.

NONE OF THE THINGS PROPOSED HEREIN CHANGES ANY SEMANTICS OF THE SPEC WHATSOEVER.

√

PR #340, opened 2017-02-17
TCK: make §3.9 tests optional

akarnokd:
This PR changes the TCK check for rule §3.9

> While the `Subscription` is not cancelled, `Subscription.request(long n)` MUST signal `onError` with a `java.lang.IllegalArgumentException` if the `argument is <= 0`. The cause message MUST include a reference to this rule and/or quote the full rule.

to be optional ([related discussion](https://github.com/reactive-streams/reactive-streams-jvm/pull/339#issuecomment-280333053)).

Rationale: libraries may report such kind of request accounting bugs by other means instead of the *very* expensive synchronization required by §1.3 to ensure correct error reporting by §3.9 of which clients can't do much about anyway.

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
@@ -346,13 +346,13 @@ public abstract class IdentityProcessorVerification<T> extends WithHelperPublish
   }
 
   @Override @Test
-  public void required_spec309_requestZeroMustSignalIllegalArgumentException() throws Throwable {
-    publisherVerification.required_spec309_requestZeroMustSignalIllegalArgumentException();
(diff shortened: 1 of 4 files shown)


viktorklang:
The spec would also need to be amended to make the behavior optional.

akarnokd:
> The spec would also need to be amended to make the behavior optional.

You are currently editing the spec in #339 which would conflict.

smaldini:
@reactive-streams/contributors add approved review and we'll merge 👍

akarnokd:
After the discussion and the resulting changes done to the libraries, this change is now mute.

viktorklang:
@akarnokd To refresh my memory (and any other reader) what was the conclusion? (link will do just fine for me)

akarnokd:
The specification for §3.9 stays as a MUST and thus the associated TCK test MUST be passed.

viktorklang:
@akarnokd thanks!

Issue #321, opened 2016-04-11
2.6 refinement/explanation


anthonyvdotbe:
§2.6 says:

> A Subscriber MUST call Subscription.cancel() if it is no longer valid to the Publisher without the Publisher having signaled onError or onComplete.

To me, it's unclear what it means for a Subscriber to be "valid to the Publisher". This was [explained](https://github.com/reactive-streams/reactive-streams-jvm/issues/177#issuecomment-208214078) by @viktorklang as:

> If the "downstream" completes or errors before the "upstream". Imagine for instance a Processor which only wants the first element from its "upstream". Or a Subscriber which receives a "forbidden" value. etc.

which makes sense, but I don't get that from reading the rule. I believe there are 2 issues:
- the word "valid": I don't see why, for example, a Processor which only wants the first element from its "upstream" would be any more valid/invalid than a Processor who wants all elements
- the phrase "to the Publisher": this implies to me that there's a validity relation with the Publisher (i.e. a Subscriber can be in a state of "valid, but invalid to its Publisher")

For reference, the current wording was introduced in ec7c3bd746f9bf247183cdb2595f28d9b37047de, where it previously was:

> A `Subscriber` MUST call `Subscription.cancel()` during shutdown if it still has an active `Subscription`.

Hereby, I want to propose the following:

> A Subscriber MUST call Subscription.cancel() if it no longer has an interest in receiving signals from its Publisher and the Subscription is not considered cancelled yet.

Note that this also rephrases the last part of the rule, saying _and the Subscription is not considered cancelled yet_ instead of _without the Publisher having signaled onError or onComplete_. The former is slightly more general, as there may be other conditions under which a Subscriber must consider its Subscription cancelled.

The explanation of the rule would then give concrete examples of cases where the Subscriber would no longer have such an interest (such as those given by @viktorklang above). And it would reference §2.4 as an example of when a Subscription is considered cancelled.

viktorklang:
This rule has been explained in the rationales.

PR #342, opened 2017-03-05
Minor rewording of 2.6 to make it easier to understand.

viktorklang:
Addressing #321 with minimal changes. (No changes in spec semantics)

@reactive-streams/contributors Please review and vote.


ktoso:
Wanting to look at the changeset shows me a 500 error page; does it work for you?

akarnokd:
@ktoso Same here. I had trouble with #339 as well the past week, mostly since the Amazon outage. The patch links from the email work for me:

https://github.com/reactive-streams/reactive-streams-jvm/pull/342.patch
https://github.com/reactive-streams/reactive-streams-jvm/pull/342.diff

ktoso:
Thanks, will have a look

viktorklang:
Yeah, I have issues with GH as well, 500s all over the place.

ktoso:
Uhm, if the change is only to add the `,` then I don't think it's helping enough.
The linked issue proposes a more wordy version, however we could find a balance between the current and the new one perhaps?

How about:

> A `Subscriber` MUST call `Subscription.cancel()` when it is no longer interested (or able) in receiving signals from its `Publisher`, even without the `Publisher` having signaled `onError` or `onComplete`.

viktorklang:
not, it changes "the Publisher" to "its Publisher" and adds the comma.

ktoso:
I see, the `valid to its...` still is not completely clear though, as witnessed by that issue. We don't use that wording anywhere else either AFAICS. A subscription can be valid, but this is the single case we used it for a Subscriber.

viktorklang:
No, it is not about the Subscriber being valid, it is about the Subscription not being valid.

ktoso:
Proves my point that it's easy to confuse the "it" then though?

viktorklang:
@ktoso Very good point. See my second attempt.

ktoso:
Hmm, simple wording and I think it does indeed cover all the cases we mean here. 
The extended intent section explains what we mean well now. 

Sounds good to me 👍

README.md
```diff
@@ -139,8 +139,8 @@ public interface Subscriber<T> {
...
 | <a name="2.5">5</a>       | A `Subscriber` MUST call `Subscription.cancel()` on the given `Subscription` after an `onSubscribe` signal if it already has an active `Subscription`. |
 | [:bulb:](#2.5 "2.5 explained") | *The intent of this rule is to prevent that two, or more, separate Publishers from thinking that they can interact with the same Subscriber. Enforcing this rule means that resource leaks are prevented since extra Subscriptions will be cancelled.* |
-| <a name="2.6">6</a>       | A `Subscriber` MUST call `Subscription.cancel()` if it is no longer valid to the `Publisher` without the `Publisher` having signaled `onError` or `onComplete`. |
-| [:bulb:](#2.6 "2.6 explained") | *The intent of this rule is to establish that Subscribers cannot just throw Subscriptions away when they are no longer needed, they have to call `cancel` so that resources held by that Subscription can be safely, and timely, reclaimed.* |
+| <a name="2.6">6</a>       | A `Subscriber` MUST call `Subscription.cancel()` if the `Subscription` is no longer needed. |
```
viktorklang: One question which comes to mind is if we need to be explicit here that this rule only applies as long as it doesn't violate any other rule (i.e. not calling Subscription methods after terminal state has been received etc.)

README.md
```diff
@@ -139,8 +139,8 @@ public interface Subscriber<T> {
...
-| <a name="2.6">6</a>       | A `Subscriber` MUST call `Subscription.cancel()` if it is no longer valid to the `Publisher` without the `Publisher` having signaled `onError` or `onComplete`. |
-| [:bulb:](#2.6 "2.6 explained") | *The intent of this rule is to establish that Subscribers cannot just throw Subscriptions away when they are no longer needed, they have to call `cancel` so that resources held by that Subscription can be safely, and timely, reclaimed.* |
+| <a name="2.6">6</a>       | A `Subscriber` MUST call `Subscription.cancel()` if the `Subscription` is no longer needed. |
+| [:bulb:](#2.6 "2.6 explained") | *The intent of this rule is to establish that Subscribers cannot just throw Subscriptions away when they are no longer needed, they have to call `cancel` so that resources held by that Subscription can be safely, and timely, reclaimed. An example of this would be a Subscriber which is only interested in a specific element, which would then cancel its Subscription to signal its completion to the Publisher.* |
```
viktorklang: Is this example/clarification good enough?

viktorklang:

viktorklang:
@reactive-streams/contributors Ping!

PR #346, opened 2017-03-14
Fix missing cancel() in tests that don't consume the entire source

akarnokd:
Some of the `PublisherVerification` tests didn't cancel the `Subscription` when they stopped consuming the source.

> 2.6: A Subscriber MUST call `Subscription.cancel()` if it is no longer valid to the `Publisher` without the `Publisher` having signaled `onError` or `onComplete`.
> :bulb:	The intent of this rule is to establish that `Subscriber`s **cannot just throw `Subscriptions` away** when they are no longer needed, they have to call `cancel` so that resources held by that `Subscription` can be safely, and timely, reclaimed.

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
@@ -212,17 +212,20 @@ public abstract class PublisherVerification<T> implements PublisherVerificationR
       public void run(Publisher<T> pub) throws InterruptedException {
 
         ManualSubscriber<T> sub = env.newManualSubscriber(pub);
+        try {
+            sub.expectNone(String.format("Publisher %s produced value before the first `request`: ", pub));
+            sub.request(1);
+            sub.nextElement(String.format("Publisher %s produced no element after first `request`", pub));
+            sub.expectNone(String.format("Publisher %s produced unrequested: ", pub));
+    
+            sub.request(1);
+            sub.request(2);
+            sub.nextElements(3, env.defaultTimeoutMillis(), String.format("Publisher %s produced less than 3 elements after two respective `request` calls", pub));
 
-        sub.expectNone(String.format("Publisher %s produced value before the first `request`: ", pub));
-        sub.request(1);
-        sub.nextElement(String.format("Publisher %s produced no element after first `request`", pub));
-        sub.expectNone(String.format("Publisher %s produced unrequested: ", pub));
-
-        sub.request(1);
-        sub.request(2);
-        sub.nextElements(3, env.defaultTimeoutMillis(), String.format("Publisher %s produced less than 3 elements after two respective `request` calls", pub));
-
-        sub.expectNone(String.format("Publisher %sproduced unrequested ", pub));
+            sub.expectNone(String.format("Publisher %sproduced unrequested ", pub));
+        } finally {
+            sub.cancel();
+        }
       }
     });
   }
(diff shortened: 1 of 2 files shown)


tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
```diff
@@ -212,17 +212,20 @@ public void required_spec101_subscriptionRequestMustResultInTheCorrectNumberOfPr
...
+            sub.expectNone(String.format("Publisher %s produced unrequested: ", pub));
+    
+            sub.request(1);
+            sub.request(2);
+            sub.nextElements(3, env.defaultTimeoutMillis(), String.format("Publisher %s produced less than 3 elements after two respective `request` calls", pub));
+
+            sub.expectNone(String.format("Publisher %sproduced unrequested ", pub));
+        } finally {
+            sub.cancel();
+        }
```
ktoso: good

tck/src/test/java/org/reactivestreams/tck/RangePublisherTest.java
```diff
@@ -0,0 +1,176 @@
...
+                for (Object o : stacks.get(t.getKey())) {
+                    b.append("\r\nat ").append(o);
+                }
+                fail = true;
+            }
+        }
+        if (fail) {
+            throw new AssertionError("Cancellations were missing:" + b);
+        }
+    }
```
ktoso: Useful, thanks

tck/src/test/java/org/reactivestreams/tck/RangePublisherTest.java
```diff
@@ -0,0 +1,176 @@
...
+    
+    public RangePublisherTest() {
+        super(new TestEnvironment(25));
```
ktoso: the default one should be fine, shouldn't it?

akarnokd: We use 25 ms for most of the synchronous tests in RxJava because otherwise it takes minutes to work through the 100+ TCK tests. Since this Publisher is synchronous, there is little to no chance for a timeout in my opinion.

ktoso:
TCK fixes look correct, thanks for catching this.

tck/src/test/java/org/reactivestreams/tck/RangePublisherTest.java
ktoso: This isn't the Rx suite nor does it impact the Rx suite, so that does not really matter :-) 

It does, however, break the configurability of the timeouts for how we use them in this repo: since the setting is given via env value: https://github.com/reactive-streams/reactive-streams-jvm/blob/master/.travis.yml#L14 and settings in code win over settings from env (documented here: https://github.com/reactive-streams/reactive-streams-jvm/tree/master/tck#timeout-configuration ).

Please note that timeout does not hurt suite time all that much, the more painful one is  (as explained here: https://github.com/reactive-streams/reactive-streams-jvm/tree/master/tck#timeout-configuration
 ) contributes much more to "actually waiting doing nothing time", so if anything the other one is more influential AFAIR.

Please stick to the default here so the `.travis.yml` configuration continues to work in the project, thanks.

akarnokd:
Updated to default `TestEnvironment()`.

smaldini:
:+1

tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
viktorklang: :+1:

tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
```diff
@@ -486,30 +489,39 @@ public void required_spec109_mustIssueOnSubscribeForNonNullSubscriber() throws T
...
+          env.verifyNoAsyncErrorsNoDelay();
+        } finally {
+          Subscription s = cancel.getAndSet(null);
+          if (s != null) {
+            s.cancel();
           }
-        });
-        onSubscribeLatch.expectClose("Should have received onSubscribe");
-        env.verifyNoAsyncErrorsNoDelay();
+        }
```
viktorklang: :+1:

tck/src/test/java/org/reactivestreams/tck/RangePublisherTest.java
viktorklang: Nice work, @akarnokd!

tck/src/test/java/org/reactivestreams/tck/RangePublisherTest.java
```diff
@@ -0,0 +1,176 @@
...
+            
+            int ids = id.incrementAndGet();
+            
+            RangeSubscription parent = new RangeSubscription(s, ids, start, start + count);
+            stacks.put(ids, stacktrace);
+            states.put(ids, false);
+            s.onSubscribe(parent);
+        }
+        
+        static final class RangeSubscription extends AtomicLong implements Subscription {
```
viktorklang: Would it be of greater utility to have it be an IteratorSubscription?

akarnokd: The main purpose was to track cancellation/completion in response to the TCK and this was the simplest multi-valued `Publisher` I could think of.

Working with an iterator requires a lot of try-catches around the method calls to `Iterator` because they may throw (for example, I saw libraries that hid JDBC calls behind plain iterators).

viktorklang: @akarnokd Fair enough, we can leave that to another time. :)

akarnokd: I guess you meant an example `IteratorSubscription`. How about I post a PR to the `examples` submodule with a few basic subscription types?

viktorklang: @akarnokd No, not an example. I thought it might be iteresting for the TCK to have a sync publisher which can be supplied an Iterable :)

ktoso:
In case my previous comment was not clear enough: LGTM, thanks 😉

smaldini:
@reactive-streams/contributors let me know if we can merge (or add approved in review menu)

viktorklang:
@reactive-streams/contributors Ok, merging.

PR #348, opened 2017-04-03
TCK: Enable replay-like and delay-error like Processor impls

akarnokd:
This PR extends the TCK to support two features of `Processor` implementations:

#### Replay all

Some `Processor` implementations may replay all `onNext` items they received over their lifetime to the incoming `Subscriber`. The `required_mustRequestFromUpstreamForElementsThatHaveBeenRequestedLongAgo` however expected that a late `Subscriber` receives only the very latest `onNext` item and fail the test if any of the first two was observed. The enhancement allows observing `(x, y, z)` or `(y, z)` or `(z)`.

Note that indeed this change may let `Processor` implementations pass which were supposed to emit only `z`. My opinion is that such implementation should already have unit tests verifying its emission pattern and the TCK should verify the implementation honors the RS protocol, even if it means getting 1-3 `onNext`s before the terminal event. If this is a problem, the change could be gated behind a `IdentityProcessorVerification` overridable property method, similar to the next change.

#### Emit all onNexts before an onError

Some `Processor` implementations may relay/replay all received `onNext`s items before they relay/replay a received `onError` event. This allows valid items to be processed before the sequence terminates with an error. Since this behavior may be specified explicitly on a `Processor` implementation, the TCK's `IdentityProcessorVerification` has been extended with a `strictEventOrdering()` method (default false) that when true, the otherwise failing `required_spec104_mustCallOnErrorOnAllItsSubscribersIfItEncountersANonRecoverableError` first consumes the pending item and then verifies the error state.

Since this change requires overriding a method, existing `Processor` implementations that eagerly call `onError` should still pass or fail this test as before.

#### ReplayProcessor

To demonstrate a `Processor` implementation that can exhibit both behaviors, a custom and package-private `ReplayProcessor` implementation has been added along with its TCK tests for non-strict and strict mode. (Note that this particular processor cancels its upstream if all downstream Subscribers have cancelled before the upstream's terminal event. This behavior is uncommon as ReactiveX style `Processor`s keep running and relay new events to later `Subscriber`s after the first set is gone.)

Related: https://github.com/reactive-streams/reactive-streams-dotnet/issues/30

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
@@ -691,7 +691,7 @@ public abstract class IdentityProcessorVerification<T> extends WithHelperPublish
           // some processors could emit items from the beginning, some may
           // cache for a limited time or count, therefore
           // any of the first 3 items may appear after requesing one
-          expectAnyNextElement(sub2, Arrays.asList(x, y, z));
+          expectAnyNextElement(sub2, new HashSet<T>(Arrays.asList(x, y, z)));
 
           if (totalRequests == 3) {
             expectRequest();
@@ -704,7 +704,7 @@ public abstract class IdentityProcessorVerification<T> extends WithHelperPublish
           // sub2 may not complete because it still has pending y or z
           if (!sub2.tryExpectCompletion(env.defaultTimeoutMillis())) {
               sub2.request(1);
-              expectAnyNextElement(sub2, Arrays.asList(y, z));
+              expectAnyNextElement(sub2, new HashSet<T>(Arrays.asList(y, z)));
               
               // z may still be pending
               if (!sub2.tryExpectCompletion(env.defaultTimeoutMillis())) {
@@ -774,7 +774,7 @@ public abstract class IdentityProcessorVerification<T> extends WithHelperPublish
       }
     }
 
-    public void expectAnyNextElement(ManualSubscriber<T> sub, Collection<T> expected) throws InterruptedException {
+    public void expectAnyNextElement(ManualSubscriber<T> sub, Set<T> expected) throws InterruptedException {
       final T elem = sub.nextElement(String.format("timeout while awaiting %s", expected));
       if (!expected.contains(elem)) {
         StringBuilder sb = new StringBuilder();
--- a/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
@@ -11,6 +11,8 @@
 
 package org.reactivestreams.tck;
 
+import java.util.*;
+
 import org.reactivestreams.Processor;
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
@@ -24,9 +26,6 @@ import org.reactivestreams.tck.support.SubscriberWhiteboxVerificationRules;
 import org.reactivestreams.tck.support.PublisherVerificationRules;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
-
-import java.util.HashSet;
-import java.util.Set;
 
 public abstract class IdentityProcessorVerification<T> extends WithHelperPublisher<T>
   implements SubscriberWhiteboxVerificationRules, PublisherVerificationRules {
@@ -189,6 +188,19 @@ public abstract class IdentityProcessorVerification<T> extends WithHelperPublish
       return Long.MAX_VALUE;
   }
 
+  /**
+   * Indicates that the tested implementation keeps a strict event ordering in respect to
+   * {@code onNext} and {@code onError}.
+   * Some {@code Processor} implementation may emit all {@code onNext}s received before emitting 
+   * any {@code onError}, similar to how {@code onComplete} is usually emitted after all
+   * previous {@code onNext} items have been emitted. The default implementation returns false,
+   * indicating that {@code onError} may cut ahead and get emitted even if there are
+   * {@code onNext} events ready for consumption (via {@code request()}) by the {@code Subscriber}.
+   */
+  public boolean strictEventOrdering() {
+      return false;
+  }
+  
   ////////////////////// TEST ENV CLEANUP /////////////////////////////////////
 
   @BeforeMethod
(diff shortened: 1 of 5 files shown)


tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
```diff
@@ -738,6 +774,20 @@ public void expectNextElement(ManualSubscriber<T> sub, T expected) throws Interr
       }
     }
 
+    public void expectAnyNextElement(ManualSubscriber<T> sub, Collection<T> expected) throws InterruptedException {
```
viktorklang: Shouldn't expected be a Set?

akarnokd: `Set` extends `Collection`, plus there is no `Set.of` in pre Java-9 but there is `Arrays.asList()`. For the particular test, there are 3 elements to expect at most so `asList` is okay, for any other use, you can call it with a `Set` if you want to.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -973,6 +977,21 @@ public void expectCompletion(long timeoutMillis, String errorMsg) throws Interru
       } // else, ok
     }
 
+    public boolean tryExpectCompletion(long timeoutMillis) throws InterruptedException {
+      while (timeoutMillis-- > 0) {
```
viktorklang: This is going to be wrong as it assumes it'll take a millisecond per pass.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -973,6 +977,21 @@ public void expectCompletion(long timeoutMillis, String errorMsg) throws Interru
...
+        Optional<T> value = abq.peek();
+        if (value != null) {
+          if (!value.isDefined()) {
+            abq.poll();
+            return true;
+          }
+          return false;
+        }
+        Thread.sleep(1);
```
viktorklang: Why this ISO poll(remaining)?

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
akarnokd: Updated.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
akarnokd: I don't understand this question. This routine uses peek because in case the next received element is not a terminal indicator, that element should remain available for any of the other `next` calls that may want to verify the item.

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
viktorklang: Since this is an API I'd rather have the types be specific.

```
private final <T extends Comparable<T>> Set<T> toSet(T... ts) {
   final Set<T> set;
   if (ts.length == 0) set = Collections.emptySet();
   else {
      set = new HashSet<T>();
      for(final T t : ts)
        set.add(t);
   }
   return set;
}
```

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
viktorklang: @akarnokd if you use abq.poll(remaining) instead of abq.peek() + abp.poll() you do not need to do the sleep(1)-dance.

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
akarnokd: Indeed, this is an API that should be as enabling as possible, that means using the most general type from the type hierarchy that has a `contains` method.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
akarnokd: 1) There is no such method of `abq.poll(arg0)`: [Javadoc](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ArrayBlockingQueue.html)
2) The purpose of the method is to consume a non-defined, non-null `Optional` and return true indicating the next event in line was a terminal event. If the next event is not a terminal event, don't consume it and return false and leave up to the test what to do with that situation. In other terms, `poll()` is not good because it consumes the item unconditionally.

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
viktorklang: @akarnokd I disagree, the semantics of expected is that of a Set, which is why Set is the most general type with the correct semantics, instead of requiring a HashSet or a TreeSet etc.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
viktorklang: 1: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ArrayBlockingQueue.html#poll(long,%20java.util.concurrent.TimeUnit)
2: Good point. Seems like the API is missing a peek(timeout, unit)!

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
```diff
@@ -738,6 +774,20 @@ public void expectNextElement(ManualSubscriber<T> sub, T expected) throws Interr
       }
     }
 
+    public void expectAnyNextElement(ManualSubscriber<T> sub, Set<T> expected) throws InterruptedException {
```
akarnokd: Updated.

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
```diff
@@ -189,6 +188,19 @@ public long maxSupportedSubscribers() {
...
+  /**
+   * Indicates that the tested implementation keeps a strict event ordering in respect to
+   * {@code onNext} and {@code onError}.
+   * Some {@code Processor} implementation may emit all {@code onNext}s received before emitting 
+   * any {@code onError}, similar to how {@code onComplete} is usually emitted after all
+   * previous {@code onNext} items have been emitted. The default implementation returns false,
+   * indicating that {@code onError} may cut ahead and get emitted even if there are
+   * {@code onNext} events ready for consumption (via {@code request()}) by the {@code Subscriber}.
+   */
+  public boolean strictEventOrdering() {
```
viktorklang: I'm n ot super-convinced by the name. Isn't this more about `eagerTermination`?

viktorklang: or `shortCircuitTermination`?

viktorklang:

akarnokd:
I think such non-identity processor could have separate standalone TCK class instead.

PR #356, opened 2017-04-04
Add Javadoc explanation to the TCK test methods about what they do

akarnokd:
This PR adds some explanation about what each `PublisherVerificationRules` methods do in respect to their expectation and request/emission pattern that could help tracking down the errors in the `Publisher` under test and tries to hint about what could be wrong in that `Publisher`.

I run into situations when there were errors in my `Publisher`s but the error response didn't help much and had to look into what the test actually did (in C# it was even harder due to lack of nicely linked TCK sources) and single-stepping through is cumbersome because of the several layers of test equipment within the TCK.

Let me address some concerns upfront:

#### Why `PublisherVerificationRules` and not `PublisherVerification`?

Mainly because the Javadoc would need duplication for `IdentityProcessorVerification` which is double maintenance work.

#### Why exposing the implementation details?

Knowing the request pattern or the expected emission pattern without digging into the TCK sources helps identify mistakes in the `Publisher` much faster, if not allow separate unit tests to be built with the same properties and debug the `Publisher` through it.

#### What about `SubscriberWhiteboxVerificationRules`?

If there is interest for such kind of documentation enhacements, I can do that after this.

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
+++ b/tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
@@ -135,7 +135,8 @@ public interface PublisherVerificationRules {
    * <li>in the {@code Publisher.subscribe(Subscriber)} method, if there is an upstream {@code Publisher},
    * that {@code Publisher} is actually subscribed to,</li>
    * <li>if the {@code Publisher} implementation is able to emit an {@code onError} without requests,</li> 
-   * <li>if the {@code Publisher} is non-empty as this test requires a {@code Publisher} without items.</li> 
+   * <li>if the {@code Publisher} is non-empty as this test requires a {@code Publisher} to signal an
+   * {@code onError} eagerly.</li> 
    * </ul>
    */
   void optional_spec104_mustSignalOnErrorWhenFails() throws Throwable;
--- a/tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
+++ b/tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
@@ -327,6 +327,7 @@ public interface PublisherVerificationRules {
    * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 
    * {@link #required_validate_boundedDepthOfOnNextAndRequestRecursion()} plus 1.
    * <p>If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
    * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
    * <li>make sure the {@link #required_createPublisher1MustProduceAStreamOfExactly1Element()} and {@link #required_createPublisher3MustProduceAStreamOfExactly3Elements()} tests pass,</li>
    * <li>the implementation doesn't allow unbounded recursion when {@code request()} is called from within {@code onNext}, i.e., the lack of
@@ -352,6 +353,7 @@ public interface PublisherVerificationRules {
    * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 3.
    * <p>The post-cancellation request pattern is (1, 1, 1).
    * <p>If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
    * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
    * <li>make sure the {@link #required_createPublisher1MustProduceAStreamOfExactly1Element()} and {@link #required_createPublisher3MustProduceAStreamOfExactly3Elements()} tests pass,</li>
    * <li>the cancellation indicator flag is properly persisted (may require volatile) and checked as part of the signal emission process.</li>
@@ -364,6 +366,7 @@ public interface PublisherVerificationRules {
    * multiple times should result in no signals to be emitted and should result in an thrown exception. 
    * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 1.
    * <p>If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
    * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
    * <li>make sure the {@link #required_createPublisher1MustProduceAStreamOfExactly1Element()} and {@link #required_createPublisher3MustProduceAStreamOfExactly3Elements()} tests pass,</li>
    * <li>the cancellation indicator flag is properly persisted (may require volatile) and checked as part of the signal emission process.</li>
--- a/tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
+++ b/tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
@@ -11,46 +11,488 @@
 
 package org.reactivestreams.tck.support;
 
+import org.reactivestreams.tck.TestEnvironment;
 
 /**
  * Internal TCK use only.
  * Add / Remove tests for PublisherVerification here to make sure that they arre added/removed in the other places.
  */
 public interface PublisherVerificationRules {
+  /**
+   * Validates that the override of {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()}
+   * returns a non-negative value.
+   * @throws Exception allow checked exceptions to be thrown by the implementation of this test
+   */
   void required_validate_maxElementsFromPublisher() throws Exception;
+  /**
+   * Validates that the override of {@link org.reactivestreams.tck.PublisherVerification#boundedDepthOfOnNextAndRequestRecursion()}
+   * returns a positive value.
+   * @throws Exception allow checked exceptions to be thrown by the implementation of this test
+   */
   void required_validate_boundedDepthOfOnNextAndRequestRecursion() throws Exception;
+  /**
+   * Asks for a {@code Publisher} that should emit exaclty one item and complete (both within a
+   * timeout specified by {@link TestEnvironment#defaultTimeoutMillis()}).
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} returns zero.
+   * <p>This test mostly ensures that the {@code Publisher} implementation is actually operational.
+   * If this test fails (likely with a timeout error), the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code Publisher.subscribe(Subscriber)} method has actual implementation,</li>
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, if there is an upstream {@code Publisher},
+   * that {@code Publisher} is actually subscribed to,</li>
+   * <li>if the {@code Publisher} is part of a chain, all elements actually issue a {@code request()} call
+   * in response to the test subscriber or by default to their upstream,</li>
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, the {@code Subscriber.onSubscribe} is called
+   * as part of the preparation process (usually before subscribing to other {@code Publisher}s),</li>
+   * <li>if the {@code Publisher} implementation works for a consumer that calls {@code request(1)},</li>
+   * <li>if the {@code Publisher} implementation is able to emit an {@code onComplete} without requests.</li> 
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_createPublisher1MustProduceAStreamOfExactly1Element() throws Throwable;
+  /**
+   * Asks for a {@code Publisher} that should emit exaclty three items and complete (all within a
+   * timeout specified by {@link TestEnvironment#defaultTimeoutMillis()}).
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 3.
+   * <p>This test mostly ensures that the {@code Publisher} implementation is actually operational.
+   * If this test fails (likely with a timeout error), the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code Publisher.subscribe(Subscriber)} method has actual implementation,</li>
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, if there is an upstream {@code Publisher},
+   * that {@code Publisher} is actually subscribed to,</li>
+   * <li>if the {@code Publisher} is part of a chain, all elements actually issue a {@code request()} call
+   * in response to the test subscriber or by default to their upstream,</li>
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, the {@code Subscriber.onSubscribe} is called
+   * as part of the preparation process (usually before subscribing to other {@code Publisher}s),</li>
+   * <li>if the {@code Publisher} implementation works for a consumer that calls {@code request(1)} after consuming an item,</li>
+   * <li>if the {@code Publisher} implementation is able to emit an {@code onComplete} without requests.</li> 
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_createPublisher3MustProduceAStreamOfExactly3Elements() throws Throwable;
+  /**
+   * Asks for a {@code Publisher} that responds to a request pattern of 0 (not requesting upfront), 1, 1 and 2
+   * in a timely manner. 
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 5.
+   * <p>This test ensures that the {@code Publisher} implementation correctly responds to {@code request()} calls that in
+   * total are less than the number of elements this {@code Publisher} could emit (thus the completion event won't be emitted).
+   * <p>If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
+   * <li>make sure the {@link #required_createPublisher1MustProduceAStreamOfExactly1Element()} and {@link #required_createPublisher3MustProduceAStreamOfExactly3Elements()} tests pass,</li>
+   * <li>if the {@code Publisher} implementation actually considers the cumulative request amount it receives,</li>
+   * <li>if the {@code Publisher} doesn't lose any {@code request()} signal and the state transition from idle -&gt; emitting or emitting -&gt; keep emitting works properly.</li>
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_spec101_subscriptionRequestMustResultInTheCorrectNumberOfProducedElements() throws Throwable;
+  /**
+   * Asks for a short {@code Publisher} and verifies that requesting once and with more than the length (but bounded) results in the
+   * correct number of items to be emitted (i.e., length 3 and request 10) followed by an {@code onComplete} signal. 
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 3.
+   * <p>This test ensures that the {@code Publisher} implementation can deal with larger requests than the number of items it can produce.
+   * <p>If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
+   * <li>make sure the {@link #required_createPublisher1MustProduceAStreamOfExactly1Element()} and {@link #required_createPublisher3MustProduceAStreamOfExactly3Elements()} tests pass.</li>
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_spec102_maySignalLessThanRequestedAndTerminateSubscription() throws Throwable;
+  /**
+   * Asks for a short {@code Publisher} (i.e., length 10), repeatedly subscribes to this {@code Publisher}, requests items
+   * one by one and verifies the {@code Publisher} calls the {@code onXXX} methods non-overlappingly.
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 10.
+   * <p>Note that this test is probabilistic, that is, may not capture any concurrent invocation in a {code Publisher} implementation.
+   * Note also that this test is sensitive to cases when a {@code request()} call in {@code onSubscribe()} triggers an asynchronous
+   * call to the other {@code onXXX} methods. In contrast, the test allows synchronous call chain of 
+   * {@code onSubscribe -> request -> onNext}.
+   * <p>If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
+   * <li>make sure the {@link #required_createPublisher1MustProduceAStreamOfExactly1Element()} and {@link #required_createPublisher3MustProduceAStreamOfExactly3Elements()} tests pass,</li>
+   * <li>if a {@code request()} call from {@code onSubscribe()} could trigger an asynchronous call to {@code onNext()} and if so, make sure
+   * such {@code request()} calls are deferred until the call to {@code onSubscribe()} returns normally.</li>
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void stochastic_spec103_mustSignalOnMethodsSequentially() throws Throwable;
+  /**
+   * Verifies that if the call to {@code PublisherVerification.createErrorPublisher()} returns a non-null {@code Publisher},
+   * it calls {@code onSubscribe} exactly once followed by a single call to {@code onError()} without issuing any requests and otherwise
+   * not throwing any exception.
+   * <p>If this test fails, the following could be checked within the error {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code Publisher.subscribe(Subscriber)} method has actual implementation,</li>
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, if there is an upstream {@code Publisher},
+   * that {@code Publisher} is actually subscribed to,</li>
+   * <li>if the {@code Publisher} implementation is able to emit an {@code onError} without requests,</li> 
+   * <li>if the {@code Publisher} is non-empty as this test requires a {@code Publisher} without items.</li> 
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void optional_spec104_mustSignalOnErrorWhenFails() throws Throwable;
+  /**
+   * Asks for a short {@code Publisher} (i.e., length 3) and verifies, after requesting one by one, the sequence
+   * completes normally.
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 3.
+   * <p>Note that the tests requests 1 after the items have been received and before expecting an {@code onComplete} signal.
+   * <p>If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
+   * <li>make sure the {@link #required_createPublisher1MustProduceAStreamOfExactly1Element()} and {@link #required_createPublisher3MustProduceAStreamOfExactly3Elements()} tests pass,</li>
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_spec105_mustSignalOnCompleteWhenFiniteStreamTerminates() throws Throwable;
+  /**
+   * Asks for an empty {@code Publisher} (i.e., length 0) and verifies it completes in a timely manner.
+   * <p>Note that the tests requests 1 before expecting an {@code onComplete} signal.
+   * <p>If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
+   * <li>if the {@code Publisher} is non-empty as this test requires a {@code Publisher} without items.</li> 
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void optional_spec105_emptyStreamMustTerminateBySignallingOnComplete() throws Throwable;
+  /**
+   * Currently, this test is skipped because it is unclear this rule can be effectively checked
+   * on a {@code Publisher} instance without looking into or hooking into the implementation of it.
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void untested_spec106_mustConsiderSubscriptionCancelledAfterOnErrorOrOnCompleteHasBeenCalled() throws Throwable;
+  /**
+   * Asks for a single-element {@code Publisher} and checks if requesting after the terminal event doesn't
+   * lead to more items or terminal signals to be emitted.
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 1.
+   * <p>The tests requests 10 items upfront and 10 items after its completion.
+   * <p>If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
+   * <li>the indication for the terminal state is properly persisted and a request call can't trigger emission of more items or another
+   * terminal signal.</li>
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_spec107_mustNotEmitFurtherSignalsOnceOnCompleteHasBeenSignalled() throws Throwable;
+  /**
+   * Currently, this test is skipped, although it is possible to validate an error {@code Publisher} along
+   * the same lines as {@link #required_spec107_mustNotEmitFurtherSignalsOnceOnCompleteHasBeenSignalled()}.
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void untested_spec107_mustNotEmitFurtherSignalsOnceOnErrorHasBeenSignalled() throws Throwable;
+  /**
+   * Currently, this test is skipped as the expected behavior is currently optional: cancelling a sequence just
+   * before the terminal event may or may not signal that terminal event.
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void untested_spec108_possiblyCanceledSubscriptionShouldNotReceiveOnErrorOrOnCompleteSignals() throws Throwable;
+  /**
+   * Asks for an empty {@code Publisher} and calls and verifies if {@code onSubscribe} was called before any calls
+   * to the other {@code onXXX} methods.
+   * <p>Note that this test doesn't request anything yet an {@code onNext} is not considered as a failure.
+   * <p>If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
+   * <li>the {@code Publisher.subscribe(Subscriber)} method has actual implementation,</li>
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, if there is an upstream {@code Publisher},
+   * that {@code Publisher} is actually subscribed to,</li>
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, the {@code Subscriber.onSubscribe} is called
+   * as part of the preparation process (usually before subscribing to other {@code Publisher}s).</li>
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_spec109_mustIssueOnSubscribeForNonNullSubscriber() throws Throwable;
+  /**
+   * Currently, this test is skipped because there is no common agreement on what is to be considered a fatal exception and
+   * besides, {@code Publisher.subscribe} is only allowed throw a {@code NullPointerException} and any other
+   * exception would require looking into or hooking into the implementation of the {@code Publisher}.
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void untested_spec109_subscribeShouldNotThrowNonFatalThrowable() throws Throwable;
+  /**
+   * Asks for an empty {@code Publisher} and calls {@code subscribe} on it with {@code null} that should result in
+   * a {@code NullPointerException} to be thrown.
+   * <p>If this test fails, check if the {@code subscribe()} implementation has an explicit null check (or a method dereference
+   * on the {@code Subscriber}), especially if the incoming {@code Subscriber} is wrapped or stored to be used later.
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_spec109_subscribeThrowNPEOnNullSubscriber() throws Throwable;
+  /**
+   * Verifies that if the call to {@code PublisherVerification.createErrorPublisher()} returns a non-null {@code Publisher},
+   * it calls {@code onSubscribe} exactly once followed by a single call to {@code onError()} without issuing any requests.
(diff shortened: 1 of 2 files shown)


smaldini:
Nice suff , wdyt @reactive-streams/contributors

DougLea:
Looks OK to me.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * Validates that the override of {@link org.reactivestreams.tck.PublisherVerification#boundedDepthOfOnNextAndRequestRecursion()}
+   * returns a positive value.
+   * @throws Exception allow checked exceptions to be thrown by the implementation of this test
+   */
   void required_validate_boundedDepthOfOnNextAndRequestRecursion() throws Exception;
+  /**
+   * Asks for a {@code Publisher} that should emit exaclty one item and complete (both within a
+   * timeout specified by {@link org.reactivestreams.tck.TestEnvironment#defaultTimeoutMillis()}).
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} returns zero.
+   * <p>This test mostly ensures that the {@code Publisher} implementation is actually operational.
```
viktorklang: This sentence is very vague, don't you agree?

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, the {@code Subscriber.onSubscribe} is called
+   * as part of the preparation process (usually before subscribing to other {@code Publisher}s),</li>
+   * <li>if the {@code Publisher} implementation works for a consumer that calls {@code request(1)},</li>
+   * <li>if the {@code Publisher} implementation is able to emit an {@code onComplete} without requests.</li> 
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_createPublisher1MustProduceAStreamOfExactly1Element() throws Throwable;
+  /**
+   * Asks for a {@code Publisher} that should emit exaclty three items and complete (all within a
```
viktorklang: exactly

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * timeout specified by {@link org.reactivestreams.tck.TestEnvironment#defaultTimeoutMillis()}).
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 3.
+   * <p>This test mostly ensures that the {@code Publisher} implementation is actually operational.
```
viktorklang: Too vague IMO. (Doesn't inspire confidence)

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * <p>If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
+   * <li>if the {@code Publisher} is non-empty as this test requires a {@code Publisher} without items.</li> 
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void optional_spec105_emptyStreamMustTerminateBySignallingOnComplete() throws Throwable;
+  /**
+   * Currently, this test is skipped because it is unclear this rule can be effectively checked
```
viktorklang: Great explanation, thanks!

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+  /**
+   * Currently, this test is skipped as the expected behavior is currently optional: cancelling a sequence just
+   * before the terminal event may or may not signal that terminal event.
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void untested_spec108_possiblyCanceledSubscriptionShouldNotReceiveOnErrorOrOnCompleteSignals() throws Throwable;
+  /**
+   * Asks for an empty {@code Publisher} and calls and verifies if {@code onSubscribe} was called before any calls
+   * to the other {@code onXXX} methods.
+   * <p>Note that this test doesn't request anything yet an {@code onNext} is not considered as a failure.
```
viktorklang: This sentence reads weirdly, clarify?

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * <li>make sure the {@link #required_createPublisher1MustProduceAStreamOfExactly1Element()} and {@link #required_createPublisher3MustProduceAStreamOfExactly3Elements()} tests pass,</li>
+   * <li>the implementation doesn't allow unbounded recursion when {@code request()} is called from within {@code onNext}, i.e., the lack of
+   * reentrant-safe state machine around the request amount (such as a for loop with a bound on the parameter {@code n} that calls {@code onNext}).
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_spec303_mustNotAllowUnboundedRecursion() throws Throwable;
+  /**
+   * Currently, this test is skipped because a {@code request} could enter into a synchronous computation via {@code onNext}
+   * legally and otherwise there is no common agreement on what constitutes as heavy computation.
```
viktorklang: There is a definition for "heavy computation" part of the spec now in the glossary?

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   */
   void required_spec307_afterSubscriptionIsCancelledAdditionalCancelationsMustBeNops() throws Throwable;
+  /**
+   * Asks for a short {@code Publisher} (length 10) and issues a {@code request(0)} which should trigger an {@code onError} call
+   * with an {@code IllegalArgumentException} and the message containing the string "3.9" (reference to the rule number).
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 10.
+   * <p>Note that this test expects the {@code IllegalArgumentException} being signalled through {@code onError}, not by
+   * throwing from {@code request()} (which is also forbidden) or signalling the error by any other means (i.e., through the
+   * Thread.currentThread().getUncaughtExceptionHandler() for example).
+   * <p>Note also that requesting and emission may happen concurrently and honoring this rule may require extra serialization within
```
viktorklang: "serialization" already has a bit of definition baggage in the JDK. I'd use "coordination" instead.

akarnokd:
> I trust you have generated the docs to verify the formatting etc of the outputted html?

No, just looked at the result from the Eclipse's Javadoc view.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
akarnokd: The sentence is totally fine. If the non-empty Publisher fails this test, there is a fundamental mistake in its implementation. See the hints below this javadoc line.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
akarnokd: How about rewording as "and otherwise there is no common agreemend how to detect such heavy computation reliably."

akarnokd:
Fixed spelling errors, clarified a couple of sentences.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
viktorklang: What I meant was that "mostly ensures" and "actually operational" are vague. :)

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
akarnokd: It is supposed to mean that in case the Publisher doesn't respect backpressure, this test will still pass. Otherwise the implementation is odd because without requesting, there is no point in verifying onNext has been called after an onSubscribe.

viktorklang: I think it is the "anything yet an" part, is there a comma or something missing?

akarnokd: I'll rephrase it with "anything, however, an ".

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
akarnokd: Removed those two sentences.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
akarnokd: Updated.

viktorklang:
@ktoso Yay or nay?

ktoso:
Moved continents in the last weeks, should be online this week and back to reviewing / contributing, thanks for your patience. Reading this one.

viktorklang:
@ktoso Thanks Konrad, it's not the end of the world if we need to postpone the 1.0.1 another week.

ktoso:
Started reviewing, I think not all comments are accurate, I will have to really focus on this one - review should come up during my day today I hope.

ktoso:
More comments coming now in a batch, however I think since if we're adding such comments, we should also transform the `// Verifies rule: https://github.com/reactive-streams/reactive-streams-jvm#1.1` comments into those javadoc comments so they link back to the specification.

How about adding that as first sentence in all of those javadoc comments (then a <p>)?

akarnokd:
> How about adding that as first sentence in all of those javadoc comments

I find those rule explanations not really helpful when considering the concrete unit tests. Perhaps we can link them as a second or later sentence.

ktoso:
It is the spec after all, the source of truth - we should link to it. People should not just rely on the TCK comments, but get an understanding of the spec in which these comments can (and will) help :-)

You're right about them not being the most important thing though when reading the comments - how about we add them as last sentence - with a <p> so it's nicely separated etc.

 // finishing review now, overall very good - some nitpicks

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
-
 /**
  * Internal TCK use only.
  * Add / Remove tests for PublisherVerification here to make sure that they arre added/removed in the other places.
  */
 public interface PublisherVerificationRules {
+  /**
+   * Validates that the override of {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()}
+   * returns a non-negative value.
+   * @throws Exception allow checked exceptions to be thrown by the implementation of this test
```
akarnokd: I get a lot of missing tag warnings without these.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * returns a positive value.
+   * @throws Exception allow checked exceptions to be thrown by the implementation of this test
+   */
   void required_validate_boundedDepthOfOnNextAndRequestRecursion() throws Exception;
+  /**
+   * Asks for a {@code Publisher} that should emit exactly one item and complete (both within a
+   * timeout specified by {@link org.reactivestreams.tck.TestEnvironment#defaultTimeoutMillis()})
+   * in response to a request(1).
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} returns zero.
+   * If this test fails (likely with a timeout error), the following could be checked within the {@code Publisher} implementation:
```
akarnokd: Okay.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * <ul>
+   * <li>the {@code Publisher.subscribe(Subscriber)} method has actual implementation,</li>
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, if there is an upstream {@code Publisher},
+   * that {@code Publisher} is actually subscribed to,</li>
+   * <li>if the {@code Publisher} is part of a chain, all elements actually issue a {@code request()} call
+   * in response to the test subscriber or by default to their upstream,</li>
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, the {@code Subscriber.onSubscribe} is called
+   * as part of the preparation process (usually before subscribing to other {@code Publisher}s),</li>
+   * <li>if the {@code Publisher} implementation works for a consumer that calls {@code request(1)},</li>
+   * <li>if the {@code Publisher} implementation is able to emit an {@code onComplete} without requests.</li>
```
akarnokd: Okay, but my experience is that it is rarely the case of emitting more than the requested amount but instead failing to emit or losing signal somewhere in the chain.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, if there is an upstream {@code Publisher},
+   * that {@code Publisher} is actually subscribed to,</li>
+   * <li>if the {@code Publisher} is part of a chain, all elements actually issue a {@code request()} call
+   * in response to the test subscriber or by default to their upstream,</li>
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, the {@code Subscriber.onSubscribe} is called
+   * as part of the preparation process (usually before subscribing to other {@code Publisher}s),</li>
+   * <li>if the {@code Publisher} implementation works for a consumer that calls {@code request(1)},</li>
+   * <li>if the {@code Publisher} implementation is able to emit an {@code onComplete} without requests.</li> 
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
```
akarnokd: Okay.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   */
   void required_createPublisher1MustProduceAStreamOfExactly1Element() throws Throwable;
+  /**
+   * Asks for a {@code Publisher} that should emit exactly three items and complete (all within a
+   * timeout specified by {@link org.reactivestreams.tck.TestEnvironment#defaultTimeoutMillis()}).
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 3.
```
akarnokd: Okay

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * <p>The tests requests one-by-one and verifies each single response item arrives in time.
+   * <p>If this test fails (likely with a timeout error), the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code Publisher.subscribe(Subscriber)} method has actual implementation,</li>
```
akarnokd: Maybe it's just me but 30% of the time I forget to fill in the auto-generated `subscribe()` method of `Publisher` after I've implemented the intermediate `Subscriber`. Usually a big green `// TODO auto generated method stub` looks at me once I realize the TCK failed.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, if there is an upstream {@code Publisher},
+   * that {@code Publisher} is actually subscribed to,</li>
+   * <li>if the {@code Publisher} is part of a chain, all elements actually issue a {@code request()} call
+   * in response to the test subscriber or by default to their upstream,</li>
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, the {@code Subscriber.onSubscribe} is called
+   * as part of the preparation process (usually before subscribing to other {@code Publisher}s),</li>
+   * <li>if the {@code Publisher} implementation works for a consumer that calls {@code request(1)} after consuming an item,</li>
```
akarnokd: Okay.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * <li>if the {@code Publisher} implementation is able to emit an {@code onComplete} without requests.</li>
```
akarnokd: I was in an odd position because it was declared by one of you that `onComplete` must happen without requesting yet this test requests. Thus two options here: 1) leave this and fix the test or 2) change the wording and have a rule inconsistency.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * Asks for a {@code Publisher} that responds to a request pattern of 0 (not requesting upfront), 1, 1 and 2
+   * in a timely manner. 
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 5.
+   * <p>This test ensures that the {@code Publisher} implementation correctly responds to {@code request()} calls that in
+   * total are less than the number of elements this {@code Publisher} could emit (thus the completion event won't be emitted).
+   * <p>If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
+   * <li>make sure the {@link #required_createPublisher1MustProduceAStreamOfExactly1Element()} and {@link #required_createPublisher3MustProduceAStreamOfExactly3Elements()} tests pass,</li>
+   * <li>if the {@code Publisher} implementation actually considers the cumulative request amount it receives,</li>
```
akarnokd: Okay.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
+   * <li>make sure the {@link #required_createPublisher1MustProduceAStreamOfExactly1Element()} and {@link #required_createPublisher3MustProduceAStreamOfExactly3Elements()} tests pass,</li>
+   * <li>if a {@code request()} call from {@code onSubscribe()} could trigger an asynchronous call to {@code onNext()} and if so, make sure
+   * such {@code request()} calls are deferred until the call to {@code onSubscribe()} returns normally.</li>
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void stochastic_spec103_mustSignalOnMethodsSequentially() throws Throwable;
+  /**
+   * Verifies that if the call to {@code PublisherVerification.createErrorPublisher()} returns a non-null {@code Publisher},
```
akarnokd: Okay, I'll rephrase it into the usual "Asks for ..."

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * it calls {@code onSubscribe} exactly once followed by a single call to {@code onError()} without issuing any requests and otherwise
+   * not throwing any exception.
```
akarnokd: Due to rule §2.13:

> In the case that this rule is violated, any associated Subscription to the Subscriber MUST be considered as cancelled, and the caller MUST raise this error condition in **a fashion that is adequate for the runtime environment**. 

In a more general sense, there some libraries define fatal errors, such as `OutOfMemoryError` and throw/rethrow them even from the `onXXX`. I wanted to express here that signalling the error condition in any other means will fail this test.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * <p>If this test fails, the following could be checked within the error {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code Publisher.subscribe(Subscriber)} method has actual implementation,</li>
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, if there is an upstream {@code Publisher},
+   * that {@code Publisher} is actually subscribed to,</li>
+   * <li>if the {@code Publisher} implementation is able to emit an {@code onError} without requests,</li>
```
akarnokd: I think the more common pitfall is the coupling the emission of `onError` to a request. Plus the first sentence expresses the need for `onSubscribe`.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * Verifies that if the call to {@code PublisherVerification.createErrorPublisher()} returns a non-null {@code Publisher},
+   * it calls {@code onSubscribe} exactly once followed by a single call to {@code onError()} without issuing any requests and otherwise
+   * not throwing any exception.
+   * <p>If this test fails, the following could be checked within the error {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code Publisher.subscribe(Subscriber)} method has actual implementation,</li>
+   * <li>in the {@code Publisher.subscribe(Subscriber)} method, if there is an upstream {@code Publisher},
+   * that {@code Publisher} is actually subscribed to,</li>
+   * <li>if the {@code Publisher} implementation is able to emit an {@code onError} without requests,</li> 
+   * <li>if the {@code Publisher} is non-empty as this test requires a {@code Publisher} without items.</li>
```
akarnokd: Indeed, some `Publisher` may want to emit items before emitting an `onError` thus in order to get to the error, requests must be issued. This is called a **delay-error** mode in RxJava and Reactor and is there as a choice for library users that want to process elements that were received successfully. When the TCK encounters such `Publisher`, this test times out because nobody asked for the items before the error.

I really don't want to argue about the value of this mode in contrast to the more victorian **fail-fast** mode certain "resilient" implementations are likely aiming at, thus warning the TCK user about trying to verify a `Publisher` that has some form of **delay-error** behavior.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_spec105_mustSignalOnCompleteWhenFiniteStreamTerminates() throws Throwable;
+  /**
+   * Asks for an empty {@code Publisher} (i.e., length 0) and verifies it completes in a timely manner.
+   * <p>Note that the tests requests 1 before expecting an {@code onComplete} signal.
+   * <p>If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
+   * <li>if the {@code Publisher} is non-empty as this test requires a {@code Publisher} without items.</li>
```
akarnokd: Yes, regular `Publisher`s are created based on length but error publishers are not.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
akarnokd: :+1:

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * Currently, this test is skipped because it is unclear this rule can be effectively checked
+   * on a {@code Publisher} instance without looking into or hooking into the implementation of it.
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void untested_spec106_mustConsiderSubscriptionCancelledAfterOnErrorOrOnCompleteHasBeenCalled() throws Throwable;
+  /**
+   * Asks for a single-element {@code Publisher} and checks if requesting after the terminal event doesn't
+   * lead to more items or terminal signals to be emitted.
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 1.
+   * <p>The tests requests 10 items upfront and 10 items after its completion.
```
akarnokd: Okay.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   */
   void required_spec107_mustNotEmitFurtherSignalsOnceOnCompleteHasBeenSignalled() throws Throwable;
+  /**
+   * Currently, this test is skipped, although it is possible to validate an error {@code Publisher} along
+   * the same lines as {@link #required_spec107_mustNotEmitFurtherSignalsOnceOnCompleteHasBeenSignalled()}.
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void untested_spec107_mustNotEmitFurtherSignalsOnceOnErrorHasBeenSignalled() throws Throwable;
+  /**
+   * Currently, this test is skipped as the expected behavior is currently optional: cancelling a sequence just
```
akarnokd: I have ideas how to test this: for examle, ask for a 1+ publisher and cancel from onNext. Regardless, I'll update the description to reflect "our inability".

akarnokd: (cancel from onNext after receiving the Nth item).

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * before the terminal event may or may not signal that terminal event.
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void untested_spec108_possiblyCanceledSubscriptionShouldNotReceiveOnErrorOrOnCompleteSignals() throws Throwable;
+  /**
+   * Asks for an empty {@code Publisher} and calls and verifies if {@code onSubscribe} was called before any calls
```
akarnokd: Okay.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * to the other {@code onXXX} methods.
```
akarnokd: Okay.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+  /**
+   * Asks for an empty {@code Publisher} and calls {@code subscribe} on it with {@code null} that should result in
+   * a {@code NullPointerException} to be thrown.
+   * <p>If this test fails, check if the {@code subscribe()} implementation has an explicit null check (or a method dereference
+   * on the {@code Subscriber}), especially if the incoming {@code Subscriber} is wrapped or stored to be used later.
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_spec109_subscribeThrowNPEOnNullSubscriber() throws Throwable;
+  /**
+   * Verifies that if the call to {@code PublisherVerification.createErrorPublisher()} returns a non-null {@code Publisher},
```
akarnokd: Okay.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
   void required_spec109_mayRejectCallsToSubscribeIfPublisherIsUnableOrUnwillingToServeThemRejectionMustTriggerOnErrorAfterOnSubscribe() throws Throwable;
+  /**
+   * Currently, this test is skipped because enforcing rule §1.10 requires unlimited retention and reference-equal checks on
+   * all incoming {@code Subscriber} which is generally infeasible, plus reusing the same {@code Subscriber} instance is
+   * better detected (or ignored) inside {@code Subscriber.onSubscribe} when the method is called multiple times. 
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void untested_spec110_rejectASubscriptionRequestIfTheSameSubscriberSubscribesTwice() throws Throwable;
+  /**
+   * Ask for a single-element {@code Publisher} and subscribes to it twice, without consuming with either {@code Subscriber} instance
```
akarnokd: Okay.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * (i.e., no requests are issued).
+   * <p>Note that this test ignores what signals the {@code Publisher} emits. Any exception thrown through non-regular
```
akarnokd: :+1:

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
   void optional_spec111_maySupportMultiSubscribe() throws Throwable;
+  /**
+   * Asks for a short {@code Publisher} (length 5), subscribes 3 {@code Subscriber}s to it, requests with different
+   * patterns and checks if all 3 received the same events in the same order.
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 5.
+   * <p>The request pattern for the first {@code Subscriber} is (1, 1, 2, 1); for the second is (2, 3) and for the third is (3, 1, 1).
+   * <p>Note that this test requires a {@code Publisher} that always emits the same signals to any {@code Subscriber}, regardless of
+   * when they subscribe and how they request elements. I.e., a "live" {@code Publisher} emitting the current time would not pass this test.
+   * <p>Note that this test is optional and may appear skipped even if the behavior should be actually supported by the {@code Publisher},
+   * see the skip message for an indication of this.
```
akarnokd: :+1:

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * <li>if the {@code Publisher} doesn't lose any {@code request()} signal and the state transition from idle -&gt; emitting or emitting -&gt; keep emitting works properly.</li>
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void optional_spec111_multicast_mustProduceTheSameElementsInTheSameSequenceToAllOfItsSubscribersWhenRequestingManyUpfrontAndCompleteAsExpected() throws Throwable;
+  /**
+   * Asks for a short {@code Publisher} (length 6), requests several times from within {@code onSubscribe} and then requests
+   * one-by-one from {@code onNext}.
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 6.
+   * <p>The request pattern is 3 x 1 from within {@code onSubscribe} and one after each {@code onNext} signal.
```
akarnokd: Okay.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * <p>The test consumes the {@code Publisher} but otherwise doesn't verify the {@code Publisher} completes (however, it checks
+   * for errors).
+   * <p>If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
+   * <li>make sure the {@link #required_createPublisher1MustProduceAStreamOfExactly1Element()} and {@link #required_createPublisher3MustProduceAStreamOfExactly3Elements()} tests pass,</li>
+   * <li>if the {@code Publisher} implementation actually considers the cumulative request amount it receives,</li>
```
akarnokd: :+1:

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   */
   void required_spec303_mustNotAllowUnboundedRecursion() throws Throwable;
+  /**
+   * Currently, this test is skipped because a {@code request} could enter into a synchronous computation via {@code onNext}
+   * legally and otherwise there is no common agreement how to detect such heavy computation reliably.
+   * @throws Exception allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void untested_spec304_requestShouldNotPerformHeavyComputations() throws Exception;
+  /**
+   * Currently, this test is skipped because there is no common agreement on what constitutes as heavy computation.
```
akarnokd: Okay

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * <ul>
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
+   * <li>make sure the {@link #required_createPublisher1MustProduceAStreamOfExactly1Element()} and {@link #required_createPublisher3MustProduceAStreamOfExactly3Elements()} tests pass,</li>
+   * <li>the cancellation indicator flag is properly persisted (may require volatile) and checked as part of the signal emission process.</li>
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_spec307_afterSubscriptionIsCancelledAdditionalCancelationsMustBeNops() throws Throwable;
+  /**
+   * Asks for a short {@code Publisher} (length 10) and issues a {@code request(0)} which should trigger an {@code onError} call
```
akarnokd: I proposed that many tests should actually try multiple lengths and there is an example PR of this proposal. In practice, a chain of operators may behave differently when the `Publisher` is known to be of length 0 or 1. 2+ is handled with the same algorithm but some peculiar concurrency bugs may appear above 10-20 because of buffers spanning multiple cache lines...

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * with an {@code IllegalArgumentException} and the message containing the string "3.9" (reference to the rule number).
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 10.
+   * <p>Note that this test expects the {@code IllegalArgumentException} being signalled through {@code onError}, not by
+   * throwing from {@code request()} (which is also forbidden) or signalling the error by any other means (i.e., through the
+   * Thread.currentThread().getUncaughtExceptionHandler() for example).
```
akarnokd: Okay.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_spec309_requestZeroMustSignalIllegalArgumentException() throws Throwable;
+  /**
+   * Asks for a short {@code Publisher} (length 10) and issues a random, negative {@code request()} call which should trigger an {@code onError} call
+   * with an {@code IllegalArgumentException} and the message containing the string "3.9" (reference to the rule number).
+   * <p>The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 10.
+   * <p>Note that this test expects the {@code IllegalArgumentException} being signalled through {@code onError}, not by
+   * throwing from {@code request()} (which is also forbidden) or signalling the error by any other means (i.e., through the
+   * Thread.currentThread().getUncaughtExceptionHandler() for example).
```
akarnokd: Added `{@code }` here as well.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,489 @@
...
+   * for this test to pass in case there is a self-contained chain of them (i.e., {@code Publisher.subscribe()} creates a chain of fresh
+   * {@code Subscriber} instances where each of them only references their downstream {@code Subscriber} thus the chain can get GC'd
+   * when the reference to the final {@code Subscriber} is dropped).
+   * </ul>
+   * @throws Throwable allow arbitrary exceptions to be thrown by the implementation of this test
+   */
   void required_spec313_cancelMustMakeThePublisherEventuallyDropAllReferencesToTheSubscriber() throws Throwable;
+  /**
+   * Asks for a short {@code Publisher} (length 3) and requests {@code Long.MAX_VALUE} from it, verifying that the
+   * {@code Publisher} emits all of its items and completes normally.
```
akarnokd: Okay.

akarnokd:
Changed the javadoc for most of the feedback from @ktoso.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
viktorklang: So for correctness: the test requires a fail-fast Publisher, not an empty one.

akarnokd: "fail-fast" and "delay-error" is not part of the Reactive-Streams vocabulary, as you pointed out not long ago, so the closest thing is an empty `Publisher`.

viktorklang: How about "a Publisher which propagates an onError signal eagerly"?

akarnokd: Updated.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
ktoso: Since these are hints we should explain the few that are likely, not just one. Keeping this one is of course good but add the one I mentioned please.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
ktoso: good

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
ktoso: Will review that one next then

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
ktoso: Sounds good, thanks

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -11,46 +11,532 @@
...
+   * {@code onSubscribe -> request -> onNext}.
+   * <p>
+   * If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * <ul>
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Publisher} has some time-delay behavior,</li>
+   * <li>make sure the {@link #required_createPublisher1MustProduceAStreamOfExactly1Element()} and {@link #required_createPublisher3MustProduceAStreamOfExactly3Elements()} tests pass,</li>
+   * <li>if a {@code request()} call from {@code onSubscribe()} could trigger an asynchronous call to {@code onNext()} and if so, make sure
+   * such {@code request()} calls are deferred until the call to {@code onSubscribe()} returns normally.</li>
+   * </ul>
+   */
```
ktoso: please do add links to the spec in all tests as I mentioned previously.
We can then also delete the `// Verifies rule: https://github.com/reactive-streams/reactive-streams-jvm#3.2` inline comments from the impls

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
akarnokd: :+1:

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
akarnokd: Okay.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
akarnokd: Updated.

akarnokd:
Yes.

viktorklang:
Merged! Massive thanks @akarnokd! :+1:

Issue #354, opened 2017-04-04
TCK: some checks could be extended to accept shorter Publishers


akarnokd:
In `PublisherVerification`, many tests require 3, 5, 6, 10 or even 20 element `Publisher`s and are not run against, for example 0 or 1 `Publisher`s (for example, 309, 317, etc).

I think these tests could be extended to support more length but I don't know what's the policy about adding more tests to `PublisherVerification` because:

- changing an interface of `PublisherVerificationRule` (although it is marked as internal) is not compatible,
- changing the `PublisherVerification` surface is not binary compatible and
- may result in bitterness as new failures require more work to be fixed in a library previously deemed verified.

viktorklang:
@akarnokd Would it be possible to select a verification based on the length of the Publisher being tested?

viktorklang:
@akarnokd Wouldn't it be possible to skip the tests which are not applicable by making them pass?

akarnokd:
Currently, if a test is not applicable the implementation throws:

```java
      throw new SkipException(String.format("Unable to run this test, as required elements nr: %d is higher than supported by given producer: %d", elements, maxElementsFromPublisher()));
```

We'd have to collect all such exceptions and rethrow a composite of them, i.e., SkipException if sub-tests were skipped or AssertionError if some tests failed + some were skipped.

The question is, what style of changes are you (and @ktoso ) willing to accept to the test classes.

viktorklang:
My personal preference are: better TCK coverage, minimal-if-any-BC-breakages, better TCK failure error messages.

@ktoso?

PR #359, opened 2017-04-05
Sub tests example

akarnokd:
Example extension to the 102 test with multiple lengths. Ongoing discussion in #354.

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
@@ -11,6 +11,23 @@
 
 package org.reactivestreams.tck;
 
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertTrue;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.lang.ref.ReferenceQueue;
+import java.lang.ref.WeakReference;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+import java.util.TreeMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
@@ -24,22 +41,6 @@ import org.reactivestreams.tck.support.PublisherVerificationRules;
 import org.testng.SkipException;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
-
-import java.io.PrintWriter;
-import java.io.StringWriter;
-import java.lang.Override;
-import java.lang.ref.ReferenceQueue;
-import java.lang.ref.WeakReference;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Random;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicReference;
-
-import static org.testng.Assert.assertEquals;
-import static org.testng.Assert.assertTrue;
 
 /**
  * Provides tests for verifying {@code Publisher} specification rules.
@@ -1141,8 +1142,7 @@ public abstract class PublisherVerification<T> implements PublisherVerificationR
                                     "which this Publisher is unable to provide (as signalled by returning Long.MAX_VALUE from `maxElementsFromPublisher()`)");
         }
       int skipped = 0;
-      List<Long> currentLength = new ArrayList<Long>();
-      List<Throwable> failures = new ArrayList<Throwable>();
+      Map<Long, Throwable> failures = new TreeMap<Long, Throwable>();
       for (long elementCount : elements) {
           if (elementCount > max) {
               skipped++;
@@ -1156,12 +1156,10 @@ public abstract class PublisherVerification<T> implements PublisherVerificationR
                   if (ex.isSkip()) {
                       skipped++;
                   } else {
-                      currentLength.add(elementCount);
-                      failures.add(ex);
+                      failures.put(elementCount, ex);
                   }
               } catch (Throwable ex) {
-                  currentLength.add(elementCount);
-                  failures.add(ex);
+                  failures.put(elementCount, ex);
               }
           }
       }
@@ -1174,9 +1172,10 @@ public abstract class PublisherVerification<T> implements PublisherVerificationR
           
           sb.append("Some of the length-specific sub-tests failed:\n");
           
-          for (int i = 0; i < currentLength.size(); i++) {
-            Throwable ex = failures.get(i);
-            sb.append("Length = ").append(currentLength.get(i)).append(": ").append(ex.getMessage()).append('\n');
+          for (Map.Entry<Long, Throwable> entry : failures.entrySet()) {
+            long len = entry.getKey();
+            Throwable ex = entry.getValue();
+            sb.append("Length = ").append(len).append(": ").append(ex.getMessage()).append('\n');
             StringWriter sw = new StringWriter();
             ex.printStackTrace(new PrintWriter(sw, true));
             sb.append(sw);
--- a/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
@@ -11,19 +11,35 @@
 
 package org.reactivestreams.tck;
 
-import static org.testng.Assert.*;
-
-import java.io.*;
-import java.lang.ref.*;
-import java.util.*;
-import java.util.concurrent.atomic.*;
-
-import org.reactivestreams.*;
-import org.reactivestreams.tck.TestEnvironment.*;
-import org.reactivestreams.tck.support.*;
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+import org.reactivestreams.tck.TestEnvironment.BlackholeSubscriberWithSubscriptionSupport;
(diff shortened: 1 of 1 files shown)


tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
```diff
@@ -1108,6 +1111,86 @@ public void activePublisherTest(long elements, boolean completionSignalRequired,
...
+          return;
+      }
+      long max = maxElementsFromPublisher();
+      if (completionSignalRequired && max == Long.MAX_VALUE) {
+          throw new SkipException("Unable to run this test, as it requires an onComplete signal, " +
+                                    "which this Publisher is unable to provide (as signalled by returning Long.MAX_VALUE from `maxElementsFromPublisher()`)");
+        }
+      int skipped = 0;
+      List<Long> currentLength = new ArrayList<Long>();
+      List<Throwable> failures = new ArrayList<Throwable>();
```
viktorklang: Perhaps represent as ``TreeMap<Long, Throwable>``?

tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
```diff
@@ -1108,6 +1111,86 @@ public void activePublisherTest(long elements, boolean completionSignalRequired,
...
+      
+      if (skipped == elements.length) {
+          throw new SkipException(String.format("Unable to run this test, as all of the required elements nr: %s is higher than supported by given producer: %d", Arrays.toString(elements), maxElementsFromPublisher()));      
+      }
+      if (!failures.isEmpty()) {
+          StringBuilder sb = new StringBuilder();
+          
+          sb.append("Some of the length-specific sub-tests failed:\n");
+          
+          for (int i = 0; i < currentLength.size(); i++) {
```
viktorklang: Isn't there similar print logic added elsewhere? (perhaps time to coalesce into a reusable piece)

viktorklang:
@reactive-streams/contributors Please weigh in :)

tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
akarnokd: Updated to TreeMap

tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
```diff
@@ -1108,6 +1112,84 @@ public void activePublisherTest(long elements, boolean completionSignalRequired,
   }
 
   /**
+   * Runs the test body with multiple {@code Publisher} lengths, collects the individual failures
+   * for those lengths and throws an {@code AssertionError} if there were any failures.
+   * If the body throws a {@code SkipException} with {@code isSkip()} {@code true}, the particular length test is ignored.
```
ktoso: link to `public void notVerified(String message) {`, perhaps as `for example as thrown by notVerified()` etc

tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
```diff
@@ -1108,6 +1112,84 @@ public void activePublisherTest(long elements, boolean completionSignalRequired,
...
+          });
+          return;
+      }
+      long max = maxElementsFromPublisher();
+      if (completionSignalRequired && max == Long.MAX_VALUE) {
+          throw new SkipException("Unable to run this test, as it requires an onComplete signal, " +
+                                    "which this Publisher is unable to provide (as signalled by returning Long.MAX_VALUE from `maxElementsFromPublisher()`)");
+        }
+      int skipped = 0;
+      Map<Long, Throwable> failures = new TreeMap<Long, Throwable>();
```
ktoso: final Map

tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
```diff
@@ -1108,6 +1112,84 @@ public void activePublisherTest(long elements, boolean completionSignalRequired,
...
+      }
+      if (!failures.isEmpty()) {
+          StringBuilder sb = new StringBuilder();
+          
+          sb.append("Some of the length-specific sub-tests failed:\n");
+          
+          for (Map.Entry<Long, Throwable> entry : failures.entrySet()) {
+            long len = entry.getKey();
+            Throwable ex = entry.getValue();
+            sb.append("Length = ").append(len).append(": ").append(ex.getMessage()).append('\n');
```
ktoso: `Failure for length [...]: ...`?

ktoso:
I'm not against this change and style per se, I'm not convinced about it's actual useful ness though. I feel we're imagining we're helping but not sure we actually are.

Complexity increase is not too bad, it's localised I guess, so not against including it though would like to see at least some case where it'd actually uncover bugs. I guess it plays around the idea that longer sequences have a higher change to show some racy behaviour than short ones?

akarnokd:
Implementations may optimize for 0 or 1 length `Publisher`s, for example a `range(n, 1) == just(n)` thus when the TCK runs for the `RangePublisher`, it actually  tests the `JustPublisher` when run against a length of 1. Same could be true for `range(n, 0) == empty()`. Such optimizations usually happen in assembly-time, i.e.:

```java
@Override public void createPublisher(long n) {
    return range(1, n);
}
// in the library to be validated
public static Publisher<Integer> range(int start, int count) {
    if (count == 0) {
       return empty()
    }
    if (count == 1) {
       return new JustPublisher(start);
    }
    return new RangePublisher(start, count);
}
```

and sometimes inside the publisher itself:

```java
class RangePublisher implements Publisher<Integer> {
   
    final int start;
    final int count;

    RangePublisher(int start, int count) {
        this.start = start;
        this.count= count;
    }

    @Override public void subscribe(Subscriber<? super Integer> s) {
        if (count == 0) {
            s.onSubscribe(EmptySubscription.INSTANCE);
            s.onComplete();
            return;
        }
        if (count == 1) {
           s.onSubscribe(new ScalarSubscription(s, start));
           return;
        }
        s.onSubscribe(new RangeSubscription(s, start, count));
    }
}
```

Unlike the fusion optimizations, these can't be just hidden away but have to be "hacked out" of the library if one wan'ts to verify the operator in its fullest.

ktoso:
Right, I agree impls are likely to special case `0` and `1` element Publishers. I'm not entirely sold about having to check such wide range, just because someone might have special cased his for-example `n < 20`-element implementation

Which tests would we want to apply this technique to?
Have we checked what the run length overhead is if we apply it to all the tests that would want this style of checks?

A running-for-aeons TCK is likely to put off people from running it continuously, so it's something to be aware of. We want to avoid the scenario that an impl just once tested with the TCK, declared it's compatible, then never run it again, introduced bugs and isn't actually compatible anymore - which would show up in production when people mix libraries for example.

In this specific test for example, the most common optimisations so 0 and 1 don't really come into play - it's `n > 2` already and then one asks oneself "how deep does the rabbit hole go?" with the length of these checks - do they need to be configurable for example? I think


> Unlike the fusion optimizations, these can't be just hidden away but have to be "hacked out" of the library if one wan'ts to verify the operator in its fullest.

That's implementation details if it has to be hacked out or not though.


Since the TCK should be as helpful as possible, I'm not against this change, but let's polish it so it's definitely useful for all styles of users.

akarnokd:
> Which tests would we want to apply this technique to?

My aim was to make sure there are little to no skipped tests when verifying a `range`-like `Publisher`; that includes widening the test numbers and implementing others sensibly for shorter ones. I'm not sure I want to pursue this direction in this TCK anymore. Here are a couple of design shortcomings that I don't think can be addressed in the TCK without extensive changes:

- check as many rules for all sorts of Publisher lengths as makes sense,
- introduce minPublisherLength because some sources can't produce 0 elements,
- allow Publishers that respond with fewer elements than requested,
- test error publishers as intensively as normal publishers (implies expecting an error at the end instead of a complete),
- this latter also requires supporting both fail-fast and delay-error behaviors.


> A running-for-aeons TCK is likely to put off people from running it continuously, so it's something to be aware of.

If tests wait for events that should not happen introduce a sleep with the TestEnvironment defaults and thus testing them for multiple lengths adds a multiple of that. For example, the default 100ms x 7 will increase the test time noticeably.

However, because the defaults are already make the test wait so much with the current implementation, I'd guess many already lower the timeout (25ms for RxJava) to make it less painful.

> declared it's compatible, then never run it again

There is no guarantee this doesn't happen right now...

> it's n > 2

Longer and possibly async `Publisher`s are more likely to hang or fail because of the bad request management and/or the sizing of the internal buffers/queues. For example, there were many bugs with RxJava when the source was longer than 128 (the default buffer size) and failed to emit to consumers. These `Publisher`s all passed the TCK. Now 128 is very excessive for async sources so imposing them on any `Publisher` I'd certainly avoid. However, configuring RxJava to 4, 8 or 16 element buffers is possible. Sometimes even 1 is workable but then there are still those skipped tests due to length...

> do they need to be configurable for example

Maybe, but not sure there is only one set required. Also there is a risk that libraries override it and later tunrs out their implementation fails at length > 1.

viktorklang:
@reactive-streams/contributors Any consensus here?

ktoso:
Looking at this again, I agree that having the ability to execute with/for longer sources would be indeed good idea in general. I do think it's hard though to pull off in practice and likely not for the 1.0.1... What do you think contributors?

viktorklang:
I think it's a bit late if we want to start shipping 1.0.1

Is there any estimate on when we could have this feature in place?

ktoso:
What I'm saying is "would be nice, but not though anyway" if that was somehow unclear :)

viktorklang:
@ktoso Ok, let's leave it for now.

ktoso:
Remove from 1.0.1 milestone, at least for now?

akarnokd:
Due to the delicate nature of the TCK, I believe more thorough testing should be implemented in a separate library. There, the main purpose would be to focus on stress testing the implementations beyond the standard TCK compliance and perhaps have a shot at testing a couple of untested rules.

PR #362, opened 2017-04-28
SubscriberBlackboxVerificationRules explained

akarnokd:
This PR adds textual explanations to the `SubscriberBlackBoxVerificationRules` and also removes the `// Verifies rule:` comments from `SubscriberBlackBoxVerification`.

The description contains a **Notes:** section which could be considered as problems, bugs or opportunities within the TCK tests. I think most of them are spec violations by the test infrastructure itself.

I haven't looked into `SubsriberWhiteBoxVerification` implementation to see what's different from the black-box version and thus the test need individual explanations or not.

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/support/SubscriberBlackboxVerificationRules.java
+++ b/tck/src/main/java/org/reactivestreams/tck/support/SubscriberBlackboxVerificationRules.java
@@ -11,34 +11,359 @@
 
 package org.reactivestreams.tck.support;
 
+import org.reactivestreams.tck.SubscriberBlackboxVerification;
+
 /**
  * Internal TCK use only.
  * Add / Remove tests for SubscriberBlackboxVerification here to make sure that they arre added/removed in the other places.
  */
 public interface SubscriberBlackboxVerificationRules {
+  /**
+   * Asks for a {@code Subscriber} instance, expects it to call {@code request()} in 
+   * a timely manner and signals as many {@code onNext} items as the very first request 
+   * amount specified by the {@code Subscriber}.
+   * <p>
+   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#2.1'>2.1</a>
+   * <p>
+   * Notes:
+   * <ul>
+   * <li>This test emits the number of items requested thus the {@code Subscriber} implementation
+   * should not request too much.</li>
+   * <li>Only the very first {@code request} amount is considered.</li>
+   * <li>This test doesn't signal {@code onComplete} after the first set of {@code onNext} signals
+   * has been emitted and may cause resource leak in
+   * {@code Subscriber}s that expect a finite {@code Publisher}.</li>
+   * <li>The test ignores cancellation from the {@code Subscriber} and emits the requested amount regardless.</li>
+   * </ul>
+   * <p>
+   * If this test fails, the following could be checked within the {@code Subscriber} implementation:
+   * <ul>
+   * <li>if the {@code Subscriber} requires external stimulus to begin requesting; override the
+   * {@link SubscriberBlackboxVerification#triggerRequest(org.reactivestreams.Subscriber)} method
+   *  in this case,</li>
+   * <li>the {@code TestEnvironment} has large enough timeout specified in case the {@code Subscriber} has some time-delay behavior,</li>
+   * <li>if the {@code Subscriber} requests zero or a negative value in some circumstances,</li>
+   * <li>if the {@code Subscriber} throws an unchecked exception from its {@code onSubscribe} or
+   * {@code onNext} methods.
+   * </ul>
+   */
   void required_spec201_blackbox_mustSignalDemandViaSubscriptionRequest() throws Throwable;
+  /**
+   * Currently, this test is skipped because there is no agreed upon approach how 
+   * to detect if the {@code Subscriber} really goes async or just responds in
+   * a timely manner.
+   * <p>
+   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#2.2'>2.2</a>
+   */
   void untested_spec202_blackbox_shouldAsynchronouslyDispatch() throws Exception;
+  /**
+   * Asks for a {@code Subscriber}, signals an {@code onSubscribe} followed by an {@code onComplete} synchronously,
+   * and checks if neither {@code request} nor {@code cancel} was called from within the {@code Subscriber}'s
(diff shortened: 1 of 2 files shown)


viktorklang:
@reactive-streams/contributors I'm trying to wrap some other RS things up for the 1.0.1 rel3ease, and I'd love to include this, if someone would have a slot to review this that'd be just great[^1].

[^1]: Mkay

DougLea:
On a quick browse, this looks OK. Although the use of "verifies" still bothers me. "Checks" would be more appropriate.

viktorklang:
@DougLea It's "verifies" as in "tests" and not as in "formal verification" :)

tck/src/main/java/org/reactivestreams/tck/support/SubscriberBlackboxVerificationRules.java
```diff
@@ -11,34 +11,359 @@
...
+   * Currently, this test is skipped because signal production is the responsibility of the {@code Publisher} and
+   * a {@code Subscription} is not expected to be the active element in an established subscription.
+   * <p>
+   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#3.11'>3.11</a>
+   * <p>
+   * Notes:
+   * <ul>
+   * <li>Tests {@link #required_spec209_blackbox_mustBePreparedToReceiveAnOnCompleteSignalWithPrecedingRequestCall() §2.9}
+   * and {@link #required_spec210_blackbox_mustBePreparedToReceiveAnOnErrorSignalWithPrecedingRequestCall() §2.10} are
+   * supposed to cover this case from the {@code Subscriber's} perspective.</li>
```
ktoso: 👍

tck/src/main/java/org/reactivestreams/tck/support/SubscriberBlackboxVerificationRules.java
```diff
@@ -11,34 +11,359 @@
...
   void untested_spec308_blackbox_requestMustRegisterGivenNumberElementsToBeProduced() throws Throwable;
+  /**
+   * Currently, this test is skipped because element production is the responsibility of the {@code Publisher} and
+   * a {@code Subscription} is not expected to be the active element in an established subscription.
+   * <p>
+   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#3.10'>3.10</a>
+   * <p>
+   * Notes:
+   * <ul>
+   * <li>This could be tested with a synchronous source currently not available within the TCK.</li>
```
ktoso: let's not sneak in non-spec wording, so with a "synchronous publisher".
We could roll-one in place for this test, feel free to PR if you want.

tck/src/main/java/org/reactivestreams/tck/support/SubscriberBlackboxVerificationRules.java
```diff
@@ -11,34 +11,359 @@
...
   void required_spec209_blackbox_mustBePreparedToReceiveAnOnCompleteSignalWithoutPrecedingRequestCall() throws Throwable;
+  /**
+   * Asks for a {@code Subscriber}, signals {@code onSubscribe} followed by an {@code onError} synchronously.
+   * <p>
+   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#2.10'>2.10</a>
+   * <p>
+   * Notes:
+   * <ul>
+   * <li>Despite the method name, the test doesn't expect a request signal from {@code Subscriber} and emits the
+   * {@code onError} signal anyway.
```
ktoso: whoops, that's a mistake then.
Addressed it in: https://github.com/reactive-streams/reactive-streams-jvm/pull/374

We (attempt to) cover both cases now, without explicitly waiting for the request which could mean we issue before the request happens, and with explicitly waiting, which means it always is after the request signal. We use the TCK infra to force a request be made.

tck/src/main/java/org/reactivestreams/tck/support/SubscriberBlackboxVerificationRules.java
```diff
@@ -11,34 +11,359 @@
...
   void required_spec209_blackbox_mustBePreparedToReceiveAnOnCompleteSignalWithPrecedingRequestCall() throws Throwable;
+  /**
+   * Asks for a {@code Subscriber} and expects it to handle {@code onComplete} independent of whether the {@code Subscriber}
+   * requests items or not.
+   * <p>
+   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#2.9'>2.9</a>
+   * <p>
+   * Notes:
+   * <ul>
+   * <li>Currently, the test doesn't call {@code onSubscribe} on the {@code Subscriber} which violates §1.9.
```
ktoso: Ouch, good catch - must have remained since days where this was legal.

ktoso: Fixed in https://github.com/reactive-streams/reactive-streams-jvm/pull/375

akarnokd:
> Please review those (thanks @akarnokd for already doing so).

Done. Note that #374 failed because some internal type casts fail.

> This PR only needs some minor rewordings, the rest is LGTM.

Do you want this PR updated first or do you want to adjust the wording along with removing the notes?

ktoso:
Works either way for me, perhaps easier to get this in and amend as we go along merging the fixes.
Knowing how people are busy merging itself may pose a time-challange... Though I'll be poking people.

akarnokd:
> perhaps easier to get this in

Agree with this and the suggested wording changes as well.

viktorklang:
@ktoso You mean you want me to merge this as is, now?  (just making sure)

ktoso:
I'm ok with that, and then we continue swiftly merging the follow ups. I'd tag them as pick-next but lack super powers :)

viktorklang:
@ktoso Ok. I think we need to do an RC of 1.0.1 to make sure that all currently passing impls have a chance to check if the new TCK works with existing, passing, TCK. Do you agree?

ktoso:
Yes, RCs are likely a good idea since we fix a bunch of things here and there in the TCK, it also gives us time in case something is wrong to fix it.

viktorklang:
@ktoso Agreed. Would be goo to give all @reactive-streams/contributors and community a couple of weeks to try it out and report back before committing to the final 1.0.1

viktorklang:
@ktoso @akarnokd I'll merge this as soon as #374 tests are green, sounds good?

Issue #278, opened 2015-06-29
Incomplete documentation on stochastic tests in TCK


jroper:
From the code:

``` java
  /**
   * Override and return {@code true} in order to skip executing tests marked as {@code Stochastic}.
   * Such tests MAY sometimes fail even though the impl
   */
  public boolean skipStochasticTests() {
    return false;
  }
```

Even though the impl what?  Under what circumstances may they fail?  I'm never seeing them fail locally, but I see them fail quite frequently in CI, is there anything that should be documented here about the TCK in order for me to determine if the failures are expected or not?

ktoso:
Ouch, seems we've cut the comment some time during the development, thanks for noticing! I'll dig out the full comment in a bit, but it wasn't anything very precise.

The only stochastic test we have is `stochastic_spec103_mustSignalOnMethodsSequentially`, it tries to see if any of the subscriber `on*` methods is called at the same time. Can you point me to a failure and impl you're getting?

viktorklang:
@jroper Ping

viktorklang:
@jroper Annual ping ;)

PR #370, opened 2017-05-22
Removes a partial sentence in the javadoc

viktorklang:
The original issue #278 has been open for almost 2 years with no one else commenting.
Improving the status quo by removing the partial sentence for now.

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
@@ -128,7 +128,9 @@ public abstract class PublisherVerification<T> implements PublisherVerificationR
 
   /**
    * Override and return {@code true} in order to skip executing tests marked as {@code Stochastic}.
-   * Such tests MAY sometimes fail even though the impl
+   * Stochastic in this case means that the Rule is impossible or infeasible to deterministically verify—
+   * usually this means that this test case can yield false positives ("be green") even if for some case,
+   * the given implementation may violate the tested behaviour.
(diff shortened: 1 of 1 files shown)


tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
```diff
@@ -128,7 +128,6 @@ public long maxElementsFromPublisher() {
 
   /**
    * Override and return {@code true} in order to skip executing tests marked as {@code Stochastic}.
-   * Such tests MAY sometimes fail even though the impl
```
ktoso: Or simply paraphrase the docs from tck/README:

`Stochastic tests can yield false positives ("be green") even if for some case, the given implementation may violate the tested behaviour. If for some reason, a stochastic test fails for your implementation, yet you are confident the implementation is correct you can disable them using this flag. Please contact the working group on https://github.com/reactive-streams/reactive-streams-jvm in case you encounter such issue with the tests.`

ktoso: See https://github.com/reactive-streams/reactive-streams-jvm/tree/master/tck#test-types-explained

viktorklang: Good point. I'll add:

"Stochastic in this case means that the Rule is impossible or infeasible to deterministically verify— usually this means that this test case can yield false positives ("be green") even if for some case, the given implementation may violate the tested behaviour."

PR #372, opened 2017-05-29
=tck #362 signal onComplete in 201 blackbox verification

ktoso:
Follow up to things found in https://github.com/reactive-streams/reactive-streams-jvm/pull/362

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/SubscriberBlackboxVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/SubscriberBlackboxVerification.java
@@ -96,11 +96,16 @@ public abstract class SubscriberBlackboxVerification<T> extends WithHelperPublis
       @Override
       public void run(BlackboxTestStage stage) throws InterruptedException {
         triggerRequest(stage.subProxy().sub());
-        final long n = stage.expectRequest();// assuming subscriber wants to consume elements...
+        final long requested = stage.expectRequest();// assuming subscriber wants to consume elements...
+        final long signalsToEmit = Math.min(requested, 512); // protecting against Subscriber which sends ridiculous large demand
 
         // should cope with up to requested number of elements
-        for (int i = 0; i < n; i++)
+        for (int i = 0; i < signalsToEmit; i++)
           stage.signalNext();
+        
+        // we complete after `signalsToEmit` (which can be less than `requested`), 
+        // which is legal under https://github.com/reactive-streams/reactive-streams-jvm#1.2
+        stage.sendCompletion();
       }
     });
   }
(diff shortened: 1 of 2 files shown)


tck/src/main/java/org/reactivestreams/tck/SubscriberBlackboxVerification.java
```diff
@@ -96,11 +96,16 @@ public void required_spec201_blackbox_mustSignalDemandViaSubscriptionRequest() t
       @Override
       public void run(BlackboxTestStage stage) throws InterruptedException {
         triggerRequest(stage.subProxy().sub());
-        final long n = stage.expectRequest();// assuming subscriber wants to consume elements...
+        final long requested = stage.expectRequest();// assuming subscriber wants to consume elements...
+        final long signalsToEmit = Math.min(requested, 512); // protecting against Subscriber which sends ridiculous large demand
```
ktoso: Or could be `TestEnvironment.TEST_BUFFER_SIZE`, anything less than huge numbers is fine here though IMO

tck/src/main/java/org/reactivestreams/tck/SubscriberBlackboxVerification.java
```diff
@@ -96,11 +96,16 @@ public void required_spec201_blackbox_mustSignalDemandViaSubscriptionRequest() t
...
 
         // should cope with up to requested number of elements
-        for (int i = 0; i < n; i++)
+        for (int i = 0; i < signalsToEmit; i++)
```
akarnokd: This could also check if the `Subscriber` signalled `cancel` and not loop unnecessarily.

tck/src/main/java/org/reactivestreams/tck/SubscriberBlackboxVerification.java
```diff
@@ -96,11 +96,25 @@ public void required_spec201_blackbox_mustSignalDemandViaSubscriptionRequest() t
...
+        final long signalsToEmit = Math.min(requested, 512); // protecting against Subscriber which sends ridiculous large demand
 
         // should cope with up to requested number of elements
-        for (int i = 0; i < n; i++)
+        for (int i = 0; i < signalsToEmit && sampleIsCancelled(stage, i, 10); i++)
           stage.signalNext();
+        
+        // we complete after `signalsToEmit` (which can be less than `requested`), 
+        // which is legal under https://github.com/reactive-streams/reactive-streams-jvm#1.2
+        stage.sendCompletion();
```
viktorklang: Should we sendCompletion() even if cancelled?

ktoso: Why not? Cancellation is allowed to be "racy"

viktorklang: I guess the "polite" implementation would check if it is not cancelled before issuing the completion. (so that the implementation does not rely on onComplete if cancelled)

PR #373, opened 2017-05-29
+tck #362 complete subscriber under test once done in 205

ktoso:
Cleanup to things found in #362

### DIFF
--- a/tck/src/test/java/org/reactivestreams/tck/SubscriberBlackboxVerificationTest.java
+++ b/tck/src/test/java/org/reactivestreams/tck/SubscriberBlackboxVerificationTest.java
@@ -18,8 +18,10 @@ import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 /**
 * Validates that the TCK's {@link org.reactivestreams.tck.SubscriberBlackboxVerification} fails with nice human readable errors.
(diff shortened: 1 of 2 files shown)


tck/src/test/java/org/reactivestreams/tck/SubscriberBlackboxVerificationTest.java
```diff
@@ -112,6 +114,34 @@ public void required_spec205_blackbox_mustCallSubscriptionCancelIfItAlreadyHasAn
...
+        }
+      }
+
+      @Override
+      public void onComplete() {
+        completion.countDown();
+      }
+    }).required_spec205_blackbox_mustCallSubscriptionCancelIfItAlreadyHasAnSubscriptionAndReceivesAnotherOnSubscribeSignal();
+    
+    completion.await(1, TimeUnit.SECONDS);
```
viktorklang: Is `1s` always appropriate here?

ktoso: Why not? Usually completes almost instant, why put tight timeouts?

viktorklang: I don't know, hence the question. Since it is a "magic number" I was wondering how it was chosen. If you think it'll work fine for all impls then that's all I need :)

ktoso: Ok, yes it's enough.
It does not matter for different impls, this provides the impl and tests the tck, no different impls will be provided for this codepath.

viktorklang: Ok!

PR #374, opened 2017-05-29
+tck #362 wait for request signal in 209, and new additional tests

ktoso:
Another cleanup to things found in #362

Now we test both cases, not waiting for request and waiting for it explicitly.
Simplified setup a bit as well for existing tests in 209, while problem was pointed out about 210


ktoso:
Note from other PR:

> We (attempt to) cover both cases now, without explicitly waiting for the request which could mean we issue before the request happens, and with explicitly waiting, which means it always is after the request signal. We use the TCK infra to force a request be made.

viktorklang:
@ktoso this  fails the build:

~~~
 java.lang.ClassCastException: org.reactivestreams.tck.SubscriberWhiteboxVerification$BlackboxSubscriberProxy cannot be cast to org.reactivestreams.tck.support.SyncTriggeredDemandSubscriber
~~~

viktorklang:
Ping @ktoso

ktoso:
Will get to this during my day

ktoso:
Pushed the fix, thanks for the reminders @viktorklang !

tck/src/main/java/org/reactivestreams/tck/support/SubscriberBlackboxVerificationRules.java
```diff
@@ -216,7 +216,26 @@
...
+   * </ul>
+   */
   void required_spec210_blackbox_mustBePreparedToReceiveAnOnErrorSignalWithoutPrecedingRequestCall() throws Throwable;
+
+  /**
+   * Currently, this test is skipped because it would require analyzing what the {@code Subscriber} implementation
+   * does.
+   * <p>
+   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#2.11'>2.11</a>
+   */
```
ktoso: ah right, forgot to add the comment; thanks a ton for doing so.
Comments sound good, thanks!

viktorklang: Np! Thanks for all your hard work, @ktoso!

PR #375, opened 2017-05-29
=tck #362 blackbox 209 must issue onSubscribe before any other signal

ktoso:
Follow up to #362, the TCK violated the spec without good reason (wasn't testing for anything like that)

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/SubscriberBlackboxVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/SubscriberBlackboxVerification.java
@@ -287,7 +287,16 @@ public abstract class SubscriberBlackboxVerification<T> extends WithHelperPublis
       public void run(BlackboxTestStage stage) throws Throwable {
         final Publisher<T> pub = new Publisher<T>() {
           @Override
-          public void subscribe(Subscriber<? super T> s) {
+          public void subscribe(final Subscriber<? super T> s) {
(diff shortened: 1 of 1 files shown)


tck/src/main/java/org/reactivestreams/tck/SubscriberBlackboxVerification.java
```diff
@@ -287,7 +287,16 @@ public void required_spec209_blackbox_mustBePreparedToReceiveAnOnCompleteSignalW
       public void run(BlackboxTestStage stage) throws Throwable {
         final Publisher<T> pub = new Publisher<T>() {
           @Override
-          public void subscribe(Subscriber<? super T> s) {
+          public void subscribe(final Subscriber<? super T> s) {
+            s.onSubscribe(new Subscription() {
```
viktorklang: This is not a valid Subscription though :)

ktoso: Please be more helpful, very tired;
The TCK contains plenty "edgy" things, since they're designed to test a single thing only

viktorklang: Absolutely, I guess what we could do is to either put in a small comment explaining why, or we could throw AssertionErrors when unsupported invocations occur, or provide a more complete implementation—if we already have any in the TCK that is.

PR #376, opened 2017-06-04
optional_spec111 test addition.

egetman:
* now test verifies https://github.com/reactive-streams/reactive-streams-jvm#1.11 and
https://github.com/reactive-streams/reactive-streams-jvm#1.5 for publishers, if they support multiple subscribers.

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
@@ -573,6 +573,32 @@ public abstract class PublisherVerification<T> implements PublisherVerificationR
   }
 
   @Override @Test
+  public void optional_spec111_registeredSubscribersMustReceiveOnNextOrOnCompleteSignals() throws Throwable {
+    optionalActivePublisherTest(1, false, new PublisherTestRun<T>() {
+      @Override
+      public void run(Publisher<T> pub) throws Throwable {
+        ManualSubscriber<T> sub1 = env.newManualSubscriber(pub);
+        ManualSubscriber<T> sub2 = env.newManualSubscriber(pub);
+        //It's implementation dependant to be a multicast or unicast.
+        //So we don't know which subscriber will receive an onNext (and onComplete) and which just onComplete.
+        //Plus, even if subscription assumed to be unicast, it's implementation choice, which one will be signalled
+        //with onNext.
+        sub1.requestNextElementOrEndOfStream();
+        sub2.requestNextElementOrEndOfStream();
+        try {
+            env.verifyNoAsyncErrors();
+        } finally {
+            try {
+                sub1.cancel();
+            } finally {
+                sub2.cancel();
+            }
+        }
+      }
+    });
+  }
+
+  @Override @Test
   public void optional_spec111_multicast_mustProduceTheSameElementsInTheSameSequenceToAllOfItsSubscribersWhenRequestingOneByOne() throws Throwable {
     optionalActivePublisherTest(5, true, new PublisherTestRun<T>() { // This test is skipped if the publisher is unbounded (never sends onComplete)
       @Override
--- a/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
@@ -579,10 +579,11 @@ public abstract class PublisherVerification<T> implements PublisherVerificationR
       public void run(Publisher<T> pub) throws Throwable {
         ManualSubscriber<T> sub1 = env.newManualSubscriber(pub);
         ManualSubscriber<T> sub2 = env.newManualSubscriber(pub);
-        //It's implementation dependant to be a multicast or unicast.
-        //So we don't know which subscriber will receive an onNext (and onComplete) and which just onComplete.
-        //Plus, even if subscription assumed to be unicast, it's implementation choice, which one will be signalled
-        //with onNext.
+        // Since we're testing the case when the Publisher DOES support the optional multi-subscribers scenario,
+        // and decides if it handles them uni-cast or multi-cast, we don't know which subscriber will receive an
+        // onNext (and optional onComplete) signal(s) and which just onComplete signal.
+        // Plus, even if subscription assumed to be unicast, it's implementation choice, which one will be signalled
+        // with onNext.
         sub1.requestNextElementOrEndOfStream();
         sub2.requestNextElementOrEndOfStream();
         try {
--- a/tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
+++ b/tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
@@ -285,7 +285,8 @@ public interface PublisherVerificationRules {
    */
   void untested_spec110_rejectASubscriptionRequestIfTheSameSubscriberSubscribesTwice() throws Throwable;
   /**
-   * Ask for a single-element {@code Publisher} and subscribes to it twice, without consuming with either {@code Subscriber} instance
+   * Asks for a single-element {@code Publisher} and subscribes to it twice, without consuming with either
+   * {@code Subscriber} instance
    * (i.e., no requests are issued).
    * <p>
    * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.11'>1.11</a>
@@ -297,11 +298,12 @@ public interface PublisherVerificationRules {
    */
   void optional_spec111_maySupportMultiSubscribe() throws Throwable;
   /**
-   * Ask for a single-element {@code Publisher} and subscribes to it twice.
+   * Asks for a single-element {@code Publisher} and subscribes to it twice.
    * Each {@code Subscriber} requests for 1 element and checks if onNext or onComplete signals was received.
    * <p>
-   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.11'>1.11</a> &amp;
-   * <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.5'>1.5</a>
+   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.11'>1.11</a>,
+   * and depends on valid implementation of rule <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.5'>1.5</a>
+   * in order to verify this.
    * <p>
    * The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 1.
    * <p>
--- a/tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
+++ b/tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
@@ -297,6 +297,18 @@ public interface PublisherVerificationRules {
    */
   void optional_spec111_maySupportMultiSubscribe() throws Throwable;
   /**
+   * Ask for a single-element {@code Publisher} and subscribes to it twice.
+   * Each {@code Subscriber} requests for 1 element and checks if onNext or onComplete signals was received.
+   * <p>
+   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.11'>1.11</a> &amp;
+   * <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.5'>1.5</a>
+   * <p>
+   * The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 1.
+   * <p>
+   * Any exception thrown through non-regular means will indicate a skipped test.
+   */
+  void optional_spec111_registeredSubscribersMustReceiveOnNextOrOnCompleteSignals() throws Throwable;
+  /**
    * Asks for a short {@code Publisher} (length 5), subscribes 3 {@code Subscriber}s to it, requests with different
    * patterns and checks if all 3 received the same events in the same order.
    * <p>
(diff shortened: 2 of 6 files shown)


viktorklang:
@egetman apologies for the silence here, I've been travelling lately.
Will take a look shortly.

tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
```diff
@@ -573,6 +573,32 @@ public void run(Publisher<T> pub) throws Throwable {
   }
 
   @Override @Test
+  public void optional_spec111_registeredSubscribersMustReceiveOnNextOrOnCompleteSignals() throws Throwable {
```
viktorklang: Does this test actually verify the intent of this spec?

viktorklang:
LGTM. Second opinion, @ktoso or @reactive-streams/contributors?

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -297,6 +297,18 @@
    */
   void optional_spec111_maySupportMultiSubscribe() throws Throwable;
   /**
+   * Ask for a single-element {@code Publisher} and subscribes to it twice.
+   * Each {@code Subscriber} requests for 1 element and checks if onNext or onComplete signals was received.
+   * <p>
+   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.11'>1.11</a> &amp;
```
ktoso: I would make the & more explicit to say what it means by that,
so ", and depends on valid implementation of rule 1.5 in order to verify this"

tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
```diff
@@ -573,6 +573,32 @@ public void run(Publisher<T> pub) throws Throwable {
...
+    optionalActivePublisherTest(1, false, new PublisherTestRun<T>() {
+      @Override
+      public void run(Publisher<T> pub) throws Throwable {
+        ManualSubscriber<T> sub1 = env.newManualSubscriber(pub);
+        ManualSubscriber<T> sub2 = env.newManualSubscriber(pub);
+        //It's implementation dependant to be a multicast or unicast.
+        //So we don't know which subscriber will receive an onNext (and onComplete) and which just onComplete.
```
ktoso: Let's add a small comment that since we're testing 1.11 we can assume that either it does not support this and thing will blow up - resulting in a skipped test or that it DOES actually support multiple subscribers (so the MAY in Rule 1.11 is fulfilled). Then the sentences here are put into better context. What I mean to highlight is that it's not just the two cases that are spec wise legal, but that in THIS test we assume that it does support multiple subscribers, which means that it will do one of those things.

The test is OK, since if it fails, it will be marked as optional and skipped - however, the comment should not lead people to think directly about only those two code paths (implementors are expected to read the source sometimes).

How about we add a comment that "Since we're testing the case when the Publisher DOES support the optional multi-subscribers scenario, and decides if it handles them uni-cast or multi-cast..." and then your comment.

In general let's try to be as helpful as possible when we do such comments (perhaps not all of current ones are perfect, but let's aim for it :-))

tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
```diff
@@ -573,6 +573,32 @@ public void run(Publisher<T> pub) throws Throwable {
...
+        sub2.requestNextElementOrEndOfStream();
+        try {
+            env.verifyNoAsyncErrors();
+        } finally {
+            try {
+                sub1.cancel();
+            } finally {
+                sub2.cancel();
+            }
+        }
```
ktoso: ok :)

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -297,6 +297,18 @@
    */
   void optional_spec111_maySupportMultiSubscribe() throws Throwable;
   /**
+   * Ask for a single-element {@code Publisher} and subscribes to it twice.
+   * Each {@code Subscriber} requests for 1 element and checks if onNext or onComplete signals was received.
```
ktoso: This is a bid on the edge; imagine a Publisher that does not know if it can signal or not (if it is completed or not) until someone signals demand to it. Specifically, we request 1, it signals, we *know* wont signal more elements but the Publisher does not know that - e.g. it has to check on some external source if it can pull from it, and only does so when demand arrives - completely legal impl. So such lazy impl, may need to get an additional request(1) to actually trigger the completion.

So this test should allow both, very lazy, and eager Publisher impls to pass. Easiest is to request 2 or request 1 and then 1.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -297,6 +297,18 @@
...
+   * <p>
+   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.11'>1.11</a> &amp;
+   * <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.5'>1.5</a>
+   * <p>
+   * The test is not executed if {@link org.reactivestreams.tck.PublisherVerification#maxElementsFromPublisher()} is less than 1.
+   * <p>
+   * Any exception thrown through non-regular means will indicate a skipped test.
+   */
+  void optional_spec111_registeredSubscribersMustReceiveOnNextOrOnCompleteSignals() throws Throwable;
```
ktoso: Must be tested in https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java#L152 both the happy case and the failure case should be tested. Failure should result in a skip.

egetman:
@viktorklang @ktoso  thx for review! I'll try to fix it today.

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
egetman: @ktoso i think, that test expect such behavior...
It publisher can emit 0 elements - both subscribers will receive onComplete signal on demand.
If publisher can emit 1 element:
a) multicast - both subscribers will receive onNext signal on demand.
b) unicast - one subscriber will receive onNext signal (and optinal onComplete - impl dependent - i.e. lazy/eager publisher), and another will receive onComplete on demand (publisher already in terminal state)
If publisher can emit n > 1 elements, both subscribers will receive onNext signal on demand.

So, test verifies that signals was received, but not it's type.

ktoso: Hm, you're right - sorry was a bit tired when reviewing that :) 
Indeed seems to cover all legal cases, and failure being the skip, so I think this covers everything.

tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java
```diff
@@ -579,10 +579,11 @@ public void optional_spec111_registeredSubscribersMustReceiveOnNextOrOnCompleteS
...
         ManualSubscriber<T> sub2 = env.newManualSubscriber(pub);
-        //It's implementation dependant to be a multicast or unicast.
-        //So we don't know which subscriber will receive an onNext (and onComplete) and which just onComplete.
-        //Plus, even if subscription assumed to be unicast, it's implementation choice, which one will be signalled
-        //with onNext.
+        // Since we're testing the case when the Publisher DOES support the optional multi-subscribers scenario,
+        // and decides if it handles them uni-cast or multi-cast, we don't know which subscriber will receive an
+        // onNext (and optional onComplete) signal(s) and which just onComplete signal.
+        // Plus, even if subscription assumed to be unicast, it's implementation choice, which one will be signalled
+        // with onNext.
```
ktoso: thanks!

tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java
```diff
@@ -297,11 +298,12 @@
...
   /**
-   * Ask for a single-element {@code Publisher} and subscribes to it twice.
+   * Asks for a single-element {@code Publisher} and subscribes to it twice.
    * Each {@code Subscriber} requests for 1 element and checks if onNext or onComplete signals was received.
    * <p>
-   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.11'>1.11</a> &amp;
-   * <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.5'>1.5</a>
+   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.11'>1.11</a>,
+   * and depends on valid implementation of rule <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.5'>1.5</a>
+   * in order to verify this.
```
ktoso: thanks!

tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
```diff
@@ -717,6 +735,79 @@ public void cancel() {
...
+          }
+
+          class Demand implements Subscription {
+
+            final AtomicBoolean canceled = new AtomicBoolean();
+            Subscriber<? super Integer> subscriber;
+
+            Demand(Subscriber<? super Integer> subscriber) {
+              this.subscriber = subscriber;
+              this.subscriber.onSubscribe(this);
```
ktoso: Not a big fan of letting `this` escape before constructor has finished running - subscriber can immediately synchronously call things on us right away. I think we're pretty safe with current impl but generally would like to avoid this pattern, WDYT?

tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
```diff
@@ -717,6 +735,79 @@ public void cancel() {
...
+            if (shouldBlowUp && !demands.isEmpty()) {
+              // onSubscribe first
+              new Demand(s);
+              s.onError(new RuntimeException("Unexpected additional subscriber"));
+            } else {
+              demands.add(new Demand(s));
+            }
+          }
+
+          class Demand implements Subscription {
```
ktoso: Let's not call it demand, don't want to mislead people into getting to this class by "open class by name" accidentally and looking for "what's that demand tha these reactive streams people keep taking about" when developing an impl attempting to pass the TCK. Some boring name here would likely be better..

tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
```diff
@@ -717,6 +735,79 @@ public void cancel() {
...
+   */
+  final PublisherVerification<Integer> multiSubscribersPublisherVerification(final boolean shouldBlowUp) {
+    return new PublisherVerification<Integer>(newTestEnvironment()) {
+
+      @Override
+      public Publisher<Integer> createPublisher(final long elements) {
+        return new Publisher<Integer>() {
+
+          private final Collection<Demand> demands = new CopyOnWriteArrayList<Demand>();
+          private final AtomicLong source = new AtomicLong(elements);
```
ktoso: yup, both nice

tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
egetman: No prob, we can manually call onSubscribe after construction. 
You r right - in case of sync impl - it could be inconsistent object state.

tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
```diff
@@ -717,6 +735,79 @@ public void cancel() {
...
   /**
+   * Verification using a Publisher that supports multiple subscribers
+   * @param shouldBlowUp if true {@link RuntimeException} will be thrown during second subscription.
+   */
+  final PublisherVerification<Integer> multiSubscribersPublisherVerification(final boolean shouldBlowUp) {
+    return new PublisherVerification<Integer>(newTestEnvironment()) {
+
+      @Override
+      public Publisher<Integer> createPublisher(final long elements) {
+        return new Publisher<Integer>() {
```
viktorklang: Would it make sense to factor this implementation out and running the TCK against it?

tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
```diff
@@ -717,6 +735,79 @@ public void cancel() {
...
+                }
+              }
+            }
+
+            @Override
+            public void cancel() {
+              canceled.set(true);
+              subscriber = null;
+              // COR list, so its ok to remove demand that way
+              for (CancelableSubscription subscription : subscriptions) {
```
viktorklang: why not `subscriptions.remove(this)` ?

tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
ktoso: Who tests the impls used in the tests of the tests... I honestly think that'd be overdoing it and not providing tangible value.

The tests (of tests) are by definition full of a-bit-wrong implementations designed to *specifically* trigger a *specific failure* and that's it. I don't see how it would help in maintenance or else to have them all be fully tested and compatible otherwise.

egetman: @ktoso right - this impl just for that test, it won't pass the TCK.

viktorklang: Fair enough

tck/src/test/java/org/reactivestreams/tck/PublisherVerificationTest.java
egetman: Sorry, my mistake =)

ktoso:
Lookin good :-)

viktorklang:
@egetman Please add yourself to the Copyrightwaivers.txt in this PR as well :)

https://github.com/reactive-streams/reactive-streams-jvm/blob/master/CopyrightWaivers.txt

egetman:
done. thanks (=

viktorklang:
Thanks for your contribution to Reactive Streams TCK, @egetman!

PR #377, opened 2017-06-12
Asynchronous Stream Publisher - Example

ttulka:
Based on Supplier, for the case you don't know the count of elements
(for example streaming from a file, etc).

Publisher is controlled by the null value - enf of stream.

### DIFF
--- a/examples/src/main/java/org/reactivestreams/example/unicast/AsyncStreamPublisher.java
+++ b/examples/src/main/java/org/reactivestreams/example/unicast/AsyncStreamPublisher.java
@@ -1,0 +1,285 @@
+/************************************************************************
+* Licensed under Public Domain (CC0)                                    *
+*                                                                       *
+* To the extent possible under law, the person who associated CC0 with  *
+* this code has waived all copyright and related or neighboring         *
+* rights to this code.                                                  *
+*                                                                       *
+* You should have received a copy of the CC0 legalcode along with this  *
+* work. If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.*
+************************************************************************/
+
+package org.reactivestreams.example.unicast;
+
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
+
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+/**
+ * AsyncStreamPublisher is an implementation of Reactive Streams `Publisher`
+ * which executes asynchronously, using a provided `Executor` and produces elements
+ * from a given `Supplier` in a "unicast" configuration to its `Subscribers`.
+ * <p>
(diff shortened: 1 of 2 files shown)


examples/src/main/java/org/reactivestreams/example/unicast/AsyncStreamPublisher.java
```diff
@@ -0,0 +1,288 @@
...
+        // This handles cancellation requests, and is idempotent, thread-safe and not synchronously performing heavy computations as specified in rule 3.5
+        private void doCancel() {
+            cancelled = true;
+        }
+
+        // Instead of executing `subscriber.onSubscribe` synchronously from within `Publisher.subscribe`
+        // we execute it asynchronously, this is to avoid executing the user code (`Iterable.iterator`) on the calling thread.
+        // It also makes it easier to follow rule 1.9
+        private void doSubscribe() {
+            if (!cancelled) {
```
akarnokd: `cancelled` being true here is impossible because the only way to get that true is by submitting a `Subscription` to the `Subscriber` that happens once and under this condition.

examples/src/main/java/org/reactivestreams/example/unicast/AsyncStreamPublisher.java
```diff
@@ -0,0 +1,288 @@
...
+                // Deal with setting up the subscription with the subscriber
+                try {
+                    subscriber.onSubscribe(this);
+                } catch (final Throwable t) { // Due diligence to obey 2.13
+                    terminateDueTo(new IllegalStateException(subscriber + " violated the Reactive Streams rule 2.13 by throwing an exception from onSubscribe.", t));
```
akarnokd: The problem with crashing `onXXX` is that you may end up in an undefined state where even calling `onError` fails.

viktorklang:
Hi @ttulka!

For my understanding, does this solve a use-case that the Iterable-based example doesn't?

ttulka:
Hi @viktorklang 
yes, this example is for the case when a "hasNext()" method is not available, it means, the stream tells us when it is over by sending a 'null' value. 
A use-case could be reading lines from a file or a socket - for example using 'BufferedReader.readLine()' as the supplier.

viktorklang:
@ttulka Ah, ok, couldn't that be solved by doing a read-ahead of 1 element? (So when the `iterator()` method on `Iterable` is called we do a read-ahead and then we advance based on that?

```scala
def supplier2Iterable[T](s: Supplier[T]): Iterable[T] = new Iterable[T] {
  override def iterator(): Iterator[T] = new Iterator[T] {
    private[this] var elem = s.get()
    override def hasNext(): Boolean = elem != null
    override def next(): T =
      if (elem == null) Iterator.empty.next()
      else {
        val prev = elem
        elem = s.get()
        prev
      }
  }
```
}

ttulka:
@viktorklang This makes definitely sense, actually it is the same logic I put into the iterable publisher.
I removed my original class and created a new one - a simple extension of the iterable publisher. This is exactly the example to explain this use case.
Thank you!

PR #378, opened 2017-06-17
Be consistent in reference style

ktoso:
We use the `#.##` style in referring to rules everywhere, this one ref was using a different style - fixed that.


viktorklang:
:+1:

PR #382, opened 2017-06-27
add patriknw to CopyrightWaivers

patriknw:


ktoso:
Oh, good catch :) Thanks for remembering about that.

viktorklang:
Thanks @patriknw!

Issue #384, opened 2017-07-04
Loosen up 3.9 tests since JDK9 does not refer to RS rules directly


ktoso:
The 1.0.1-RC1 TCK is too restrictive for JDK9 to pass it since we required a reference to a rule spec in a failure. The JDK follows it's usual style of error messages and should not be expected to have to change that. 

See: https://github.com/reactive-streams/reactive-streams-jvm/issues/345#issuecomment-312911130

PR #385, opened 2017-07-04
=tck #384 dont check for cause message when checking 3.9

ktoso:
Resolves https://github.com/reactive-streams/reactive-streams-jvm/issues/384 in the TCK, existing impls will remain to pass since they were already referring to 3.9

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
@@ -370,6 +370,11 @@ public abstract class IdentityProcessorVerification<T> extends WithHelperPublish
   public void required_spec309_requestNegativeNumberMustSignalIllegalArgumentException() throws Throwable {
     publisherVerification.required_spec309_requestNegativeNumberMustSignalIllegalArgumentException();
   }
+  
+  @Override @Test
+  public void optional_spec309_requestNegativeNumberMaySignalIllegalArgumentExceptionWithSpecificMessage() throws Throwable {
+    publisherVerification.optional_spec309_requestNegativeNumberMaySignalIllegalArgumentExceptionWithSpecificMessage();
+  }
 
   @Override @Test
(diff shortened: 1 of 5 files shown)


akarnokd:
Please don't forget the TCK explanations:

[PublisherVerificationRules.java#L474](https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java#L474)

[PublisherVerificationRules.java#L499](https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/support/PublisherVerificationRules.java#L499)

ktoso:
Thanks for the reminder @akarnokd !

PR currently blocked by decision on: https://github.com/reactive-streams/reactive-streams-jvm/pull/386#discussion_r125511074 if we actually want to check the message at all or not - perhaps not.

Then I'll amend this PR and fix the comments too :)

ktoso:
Aligned with spec wording proposed in https://github.com/reactive-streams/reactive-streams-jvm/pull/386 now.

Changes are:

- we don't require any message content to pass the 3.9 rule
  - only that an IllegalArgumentException is signaled is checked  (this is compatible with old impls and JDK9)
- added an optional test to check if an impl actually references 3.9 or uses the proposed (in the spec as SHOULD) wording of "non-positive subscription request" (JDK wording) or the previous "3.9" reference which used to be required
  - this is compatible too, since it's just optional (and existing impls pass it by default since they passed the previous impl of the test)
- amended explanations of these tests - thanks for the reminder David

viktorklang:
LGTM.

@reactive-streams/contributors Please chime in so we can include this change in the RC2 ASAP. Thank you!

viktorklang:
@reactive-streams/contributors 1 day left to comment on this!

viktorklang:
@reactive-streams/contributors Merging.

Issue #345, opened 2017-03-13
Release 1.0.1


viktorklang:
I propose that we, @reactive-streams/contributors, release 1.0.1 on the 2-year anniversary of 1.0.0

*No semantically breaking changes.*

Improved TCK? Improved docs? Improved spec?

akarnokd:
How urgent is this? I might have found an issue with `PublisherVerification`: some tests don't cancel the test `Subscriber`'s `Subscription` which leaves the created `Publisher` hanging or leaking resources. 

My suspects are the `spec111_multicast_*` because in my TestNG execution, the test `required_createPublisher1MustProduceAStreamOfExactly1Element` failed when the whole suite was executed but passed when the individual test was executed.

I'm currently preparing a TCK test that should detect if `cancel` was called in a particular test if there was no onComplete from the created `Publisher`.

viktorklang:
The 2-year anniversary is on the 29th of *April* :)

smaldini:
I'll be ready to publish it but i think it's a good idea. Could we reference in this ticket issues we are going to address (or label them) ? In particular would this come with the soft breaking about request(0) not erroring ?

viktorklang:
@smaldini I'd love to discuss that "problem" (request(0)) but not on this Issue :)

I believe all tickets are assigned to the [milestone](https://github.com/reactive-streams/reactive-streams-jvm/milestone/10)

DougLea:
Just FYI, jdk was successfully changed to reject request(0) for initial jdk9 release, even though I still think it was a bad idea. But better to conform to 1.0 spec anyway...

viktorklang:
Thanks @DougLea!

viktorklang:
@reactive-streams/contributors Has everyone updated their implementations to try the new TCK?
Are we good to go on promoting the RC1 to final? Please report back here ASAP.

ktoso:
Akka Streams passing: https://github.com/akka/akka/pull/23267

akarnokd:
RxJava 2.1.2 snapshot passing: https://github.com/ReactiveX/RxJava/pull/5453

ktoso:
I made a PR in Vert.X asking to check the RC: https://github.com/vert-x3/vertx-reactive-streams/pull/9

smaldini:
I'll check in a moment today thanks @reactive-streams/contributors

viktorklang:
@smaldini Any verdict?

[...]

PR #386, opened 2017-07-04
=spec #384 amend spec to allow not mentioning rule number in exception message

ktoso:
Resolves spec part of https://github.com/reactive-streams/reactive-streams-jvm/issues/384 and is follow up / prerequisite for the TCK change: https://github.com/reactive-streams/reactive-streams-jvm/pull/385

Discussion in: https://github.com/reactive-streams/reactive-streams-jvm/issues/345#issuecomment-312917833

This change is compatible with existing implementations, and allows JDK9's interfaces to be compatible with our rules (since that impl won't refer to the number of the rule explicitly)

Resolves #385


README.md
```diff
@@ -183,7 +183,7 @@ public interface Subscription {
 | [:bulb:](#3.7 "3.7 explained") | *The intent of this rule is superseded by [3.5](#3.5).* |
 | <a name="3.8">8</a>       | While the `Subscription` is not cancelled, `Subscription.request(long n)` MUST register the given number of additional elements to be produced to the respective subscriber. |
 | [:bulb:](#3.8 "3.8 explained") | *The intent of this rule is to make sure that `request`-ing is an additive operation, as well as ensuring that a request for elements is delivered to the Publisher.* |
-| <a name="3.9">9</a>       | While the `Subscription` is not cancelled, `Subscription.request(long n)` MUST signal `onError` with a `java.lang.IllegalArgumentException` if the argument is <= 0. The cause message MUST include a reference to this rule and/or quote the full rule. |
+| <a name="3.9">9</a>       | While the `Subscription` is not cancelled, `Subscription.request(long n)` MUST signal `onError` with a `java.lang.IllegalArgumentException` if the argument is <= 0. The cause  message SHOULD explain that non-positive request signals are illegal. |
```
viktorklang: There's an extraneous space between "cause" and "message".

I guess the question is that if amending this rule should actually be removing the last sentence, since the other rules which mandate exceptions are not requiring rule referencing. Thoughts @reactive-streams/contributors ?

ktoso: Ah, I see, then in the TCK we should not check the message at all if we want to remove the wording suggestion. It would only check that we got an `IllegalArgumentException ` - which is also OK.

I don't mind either way, I like the suggestion but happy to remove and change the TCK PR to only check exception type.

ktoso: After some pondering I think just checking exception type is the most compatible thing to do anyway and to actually enforce - so I agree on removing the sentence completely.

Waiting a round of sleep for feedback and addressing those once i wake up :)

DougLea:
I think the current proposed wording is OK. 
("SHOULD explain that non-positive request signals are illegal.")
This is in keeping with the sentiment that the exception should be helpful in identifying the nature of problem vs many other possible IllegalArgumentExceptions. On the other hand "SHOULD" implies
that the actual conformance check probably shouldn't try to check for any particular wording.

ktoso:
Fixed the additional space and kept the wording, thanks for chiming in @DougLea !

viktorklang:
@DougLea @ktoso I agree. Making it a SHOULD is the most sensible option here.

viktorklang:
@reactive-streams/contributors 1 day left to comment on this!

viktorklang:
@reactive-streams/contributors Merging.

Issue #393, opened 2017-09-28
Document if Publisher.subscribe is allowed to block


guenhter:
Hi,

I'm using reactive-streams and wrote a `Publisher` for ZeroMq.
One of the first things is stumbled over was, if the method `Publisher.subscribe(...)` must return and emit the results async, or if the method can block.

I looked up the javadoc of the method, but it isn't documented. Would be great to make a statement what is expected by that method.

viktorklang:
@guenhter Did your publisher pass the TCK?

akarnokd:
I think it is unavoidable to have blocking code inside `subscribe` since we have to interact with non-reactive and often blocking code if we have to expose them as reactive in some way. Some reactive libraries have specific operators that move a possibly blocking `subscribe` call to a background thread:

```java
Flowable.fromCallable(() -> {
    Thread.sleep(1000);
    return 1;
})
.subscribeOn(Schedulers.io())
.subscribe(System.out::println);
```

The main use case is to free up the main thread. You can inline the two so your `subscribe` implementation starts a task in some thread pool which can "freely" block.

The TCK can't and doesn't detect if your `subscribe` blocks as from the TCK's perspective, they appear as synchronous sources.

viktorklang:
@akarnokd I guess it depends on the definition of "blocking" used. I think it would be surprising to users if they had to assume/do defensive backgrounding of `subscribe` calls as a general practice. I think it is fair to assume that "work is done" within subscribe, and that the method takes time to execute, but stalling  the calling thread perpetually?

akarnokd:
It happens all the time on Android with Network or file resources being streamed. Most underlying components use blocking calls (i.e., FileInputStream.read() type of blocking).

viktorklang:
@akarnokd Well, technically referencing  new Calss causes it to be loaded, which could be a URLClassLoader which could technically block forever. I'm just saying that the general case one should assume that calling `subscribe` is safe (i.e. calling thread is not blocked perpetually).

rkuhn:
@guenhter Thanks for pointing out an omission: for `Subscription.request` and `Subscription.cancel` we have documented the intention of non-obstructive behavior, perhaps that should also be added to `Publisher.subscribe`. Since all information flow between Publisher and Subscriber is asynchronous, it should not be a problem to construct a Subscription without involving blocking ømq APIs—asynchronous failures can easily be signalled using onError.

guenhter:
[...]

PR #395, opened 2017-10-03
Example synchronous range Publisher

akarnokd:
This PR adds a synchronous `Publisher` example that emits a range of values.

See #393

### DIFF
--- a/examples/src/main/java/org/reactivestreams/example/lazycast/RangePublisher.java
+++ b/examples/src/main/java/org/reactivestreams/example/lazycast/RangePublisher.java
@@ -156,7 +156,9 @@ public final class RangePublisher implements Publisher<Integer> {
 
             try {
                 for (; ; ) {
-                    // Check if there was an invalid request and then report it.
+                    // Check if there was an invalid request and then report its exception
+                    // as mandated by rule 3.9. The stacktrace in it should
+                    // help locate the faulty logic in the Subscriber.
                     Throwable invalidRequest = this.invalidRequest;
                     if (invalidRequest != null) {
                         // When we signal onError, the subscription must be considered as cancelled, as per rule 1.6
--- a/examples/src/main/java/org/reactivestreams/example/lazycast/RangePublisher.java
+++ b/examples/src/main/java/org/reactivestreams/example/lazycast/RangePublisher.java
@@ -40,8 +40,24 @@ public final class RangePublisher implements Publisher<Integer> {
     }
 
     @Override
-    public void subscribe(Subscriber<? super Integer> s) {
-        s.onSubscribe(new RangeSubscription(s, start, start + count));
+    public void subscribe(Subscriber<? super Integer> subscriber) {
+        // As per rule 1.11, we have decided to support multiple subscribers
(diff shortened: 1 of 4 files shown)


examples/src/main/java/org/reactivestreams/example/lazycast/RangePublisher.java
```diff
@@ -0,0 +1,187 @@
...
+     * @param start the starting value of the range
+     * @param count the number of items to emit
+     */
+    public RangePublisher(int start, int count) {
+        this.start = start;
+        this.count = count;
+    }
+
+    @Override
+    public void subscribe(Subscriber<? super Integer> s) {
```
viktorklang: Thanks @akarnokd!

Could you add similar documentation as https://github.com/reactive-streams/reactive-streams-jvm/blob/master/examples/src/main/java/org/reactivestreams/example/unicast/AsyncIterablePublisher.java ?

Specifically referring to Spec rules etc. (So the examples are consistent in docs)

akarnokd: Done.

examples/src/main/java/org/reactivestreams/example/lazycast/RangePublisher.java
```diff
@@ -0,0 +1,242 @@
...
+ *                                                                       *
+ * To the extent possible under law, the person who associated CC0 with  *
+ * this code has waived all copyright and related or neighboring         *
+ * rights to this code.                                                  *
+ *                                                                       *
+ * You should have received a copy of the CC0 legalcode along with this  *
+ * work. If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.*
+ ************************************************************************/
+
+package org.reactivestreams.example.lazycast;
```
viktorklang: lazycast?

akarnokd: I don't know the terminology in RS of what we call "cold observables" in RxJava.

akarnokd: Also for us, `unicast` means a maximum of 1 `Subscriber` and `multicast` is associated with `Processor`-like behavior.

akarnokd: I'm open to naming suggestion.

viktorklang: `unicast` would make most sense, since it is a sync version of the async? https://github.com/reactive-streams/reactive-streams-jvm/blob/master/examples/src/main/java/org/reactivestreams/example/unicast/AsyncIterablePublisher.java

akarnokd: Okay, I moved them there.

viktorklang:
@reactive-streams/contributors Please weigh in here. yay or nay?

viktorklang:
Thanks @akarnokd! :+1:

ktoso:
Late to the party but "Yay" :)

PR #399, opened 2017-10-12
Java 9 Flow bridge: add Subscriber converters

akarnokd:
This PR exposes the `Subscriber` conversion through the `ReactiveStreamsFlowBridge` API.

### DIFF
--- a/flow-bridge/src/main/java/org/reactivestreams/ReactiveStreamsFlowBridge.java
+++ b/flow-bridge/src/main/java/org/reactivestreams/ReactiveStreamsFlowBridge.java
@@ -110,7 +110,33 @@ public final class ReactiveStreamsFlowBridge {
         }
         return new FlowToReactiveProcessor<T, U>(reactiveStreamsProcessor);
     }
-    
+
+    /**
+     * Converts a Reactive Streams Subscriber into a Flow Subscriber.
+     * @param <T> the input and output value type
+     * @param reactiveStreamsSubscriber the Reactive Streams Subscriber instance to convert
+     * @return the equivalent Flow Subscriber
+     */
+    public static <T> Flow.Subscriber<T> toFlowSubscriber(org.reactivestreams.Subscriber<T> reactiveStreamsSubscriber) {
+        if (reactiveStreamsSubscriber == null) {
+            throw new NullPointerException("reactiveStreamsSubscriber");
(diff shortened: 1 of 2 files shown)


viktorklang:
LGTM!

viktorklang:
@reactive-streams/contributors Merge?

ktoso:
+1 :)

flow-bridge/src/main/java/org/reactivestreams/ReactiveStreamsFlowBridge.java
```diff
@@ -110,7 +110,33 @@ private ReactiveStreamsFlowBridge() {
...
+        return new FlowToReactiveSubscriber<T>(reactiveStreamsSubscriber);
+    }
+
+    /**
+     * Converts a Flow Subscriber into a Reactive Streams Subscriber.
+     * @param <T> the input and output value type
+     * @param flowSubscriber the Flow Subscriber instance to convert
+     * @return the equivalent Reactive Streams Subscriber
+     */
+    public static <T> org.reactivestreams.Subscriber<T> toReactiveStreamsSubscriber(Flow.Subscriber<T> flowSubscriber) {
```
ktoso: Hm, I wonder actually; now it's onconsistent since most methods are `toFlow` and `to ReactiveStreams` and this one adds the specific type name - should we align them to the same style?

The overloaded version works fine too I think, just `toX` everywhere, but I have no strong preference WDYT?

akarnokd: I think there might be an ambiguity problem with `toFlow(Processor)` when trying to convert a `Processor`, requiring the user to cast. For consistency though, a postfix could be added to the other  methods.

viktorklang: If the intended usage is to `import static` then I think it's fine to harmonize on "toReactiveStreamsX" and "toFlowX". Thoughts?

ktoso: Yeah good catch that a processor would be ambigious then... I'd avoid having 2 methods to do the same, so I guess for consistency we should do the long names everywhere.

PR #400, opened 2017-10-29
Wip flow tck √

viktorklang:
Dummy PR to get travis to take a look.


flow-bridge/src/main/java/org/reactivestreams/ReactiveStreamsFlowBridge.java
```diff
@@ -34,12 +34,12 @@ private ReactiveStreamsFlowBridge() {
         if (flowPublisher == null) {
             throw new NullPointerException("flowPublisher");
         }
-        if (flowPublisher instanceof org.reactivestreams.Publisher) {
-            return (org.reactivestreams.Publisher<T>)flowPublisher;
-        }
         if (flowPublisher instanceof FlowPublisherFromReactive) {
```
ktoso: By the way, before release we really should revisit naming of the classes `FlowPublisherFromReactive ` (specifically Reactive => ReactiveStreams or similar)

flow-bridge/src/main/java/org/reactivestreams/ReactiveStreamsFlowBridge.java
```diff
@@ -188,15 +202,15 @@ public void cancel() {
      */
     static final class FlowToReactiveSubscriber<T>
             implements Flow.Subscriber<T> {
-        private final org.reactivestreams.Subscriber<? super T> reactiveStreams;
+        public final org.reactivestreams.Subscriber<? super T> reactiveStreams;
```
ktoso: yeah, saw the need for exposing it, but would rather go "java style" since we're java project, so getter.
I'd call it `underlying()` or `delegate()` i guess?

flow-bridge/src/main/java/org/reactivestreams/ReactiveStreamsFlowBridge.java
```diff
@@ -188,15 +202,15 @@ public void cancel() {
...
 
         public FlowToReactiveSubscriber(org.reactivestreams.Subscriber<? super T> reactive) {
             this.reactiveStreams = reactive;
         }
 
         @Override
         public void onSubscribe(Flow.Subscription subscription) {
-            reactiveStreams.onSubscribe(new ReactiveToFlowSubscription(subscription));
+            reactiveStreams.onSubscribe((subscription == null) ? null : new ReactiveToFlowSubscription(subscription));
```
ktoso: good one

ktoso:
Lookin good, thanks so much for pushing over the final finish line

ktoso:
From my side we could squash my two commits here and merge as two commits from this PR. Thanks Viktor

viktorklang:
@ktoso 

>Lookin very good indeed; A bit worried about the exposed field, even if final hm... I wonder is "java people will panic" if you know what I mean

They won't be exposed though? (The classes are package-protected and we never return the types publically)

flow-bridge/src/main/java/org/reactivestreams/ReactiveStreamsFlowBridge.java
viktorklang: Thanks, yeah, avoiding wrapping nulls seemed like a good idea. :)

flow-bridge/src/main/java/org/reactivestreams/ReactiveStreamsFlowBridge.java
viktorklang: I can make it package-protected instead?

flow-bridge/src/main/java/org/reactivestreams/ReactiveStreamsFlowBridge.java
viktorklang: Yeah, let's do the naming discussion separately.

viktorklang:
@ktoso Should I delete the .orig files?

ktoso:
That could be good I guess indeed. One should hope that no one should have
to get into the internals of these

viktorklang:
@ktoso I can squash this into a single commit, but I don't mind having the full history for attribution, unless you have some other preference?

ktoso:
> They won't be exposed though? (The classes are package-protected and we never return the types publically)

Hmm, good point. In my work-in-progress i ended up opening up those classes; which perhaps was too early. Let’s keep the current state of what is here. If we ever need to open up we’d add the method version i guess.

ktoso:
> @ktoso I can squash this into a single commit, but I don't mind having the full history for attribution, unless you have some other preference?

No strong preference, rebasing interactive into 2 to give you credit for the fix up would be nicest I guess — “credit where it’s due” :) No preference tho.

viktorklang:
@ktoso How abotu this plan: I close this PR without merging. You squash your commits on your PR and cherry-pick https://github.com/reactive-streams/reactive-streams-jvm/pull/400/commits/03da556510f3b536782eb0f0bf2204abaf4e63dc, that way we keep the original PR, my changes are applied, and you get to control how you want your commits to be represented.

ktoso:
Ok, can do that — will be at my laptop shortly.

viktorklang:
@ktoso Awesome. I'll close this PR.

Issue #401, opened 2017-10-31
JDK9 artifact/converter naming


viktorklang:
Hey,

so we just now merged the "final bits" for the JDK9 Flow bridge/adapter jar as well as the TCK for java.util.concurrent.Flow.

Now, there are some open questions that we need to lock down now, before we cut the releases:

  1. Naming conventions on converters:
   Current scheme: toX (toSubscriber, toPublisher, toProcessor) and toFlowX (toFlowSubsriber, toFlowPublisher, toFlowProcessor)
   Current rationale: harmonizes with the types: X and Flow.X
  2. Naming of converter class.
   Current name: The converters are currently located in a class named `org.reactivestreams.ReactiveStreamsFlowBridge`
  3. Naming of converter jar
   Current artifact name is `reactive-streams-flow-bridge`
   Current rationale: it is intended to bridge implementations in both directions

@reactive-streams/contributors I'd like for us to settle the naming swiftly and with resolve (can we aim for within a week?), so if you have counter proposals, please speak now of forever hold your peace. :)

viktorklang:
One possible proposal is:

1. keep toX, toFlowX
2. Rename converter class to `org.reactivestreams.Adapt`
3. Rename converter jar to `reactive-streams-flow-adapter`

shenghaiyang:
Maybe like this? :)
```xml
<dependency>
  <groupId>org.reactivestreams</groupId>
  <artifactId>adapter-flow</artifactId>
</dependency>
```

viktorklang:
@shenghaiyang All the other artifacts start with `reactive-streams` so it would seem asymmetric to change that. :)

ktoso:
1. The `toPublisher` / `toFlowPublisher` are idea, it's good to have the name in there and that plays well with the Flow mnemonic, it's also short. The alternatives were with including "ReactiveStreams" or "RS" which is a bit weird.

2. In Java I'm usually a bit afraid to take "the good names" (`Adapt`), since conflicting in names is a pain in Java (no import renames like Scala etc). Those methods are likely to be used via a static import anyway I think, so the name does not matter too much. Sure it reads nice and fluently, but that's not really what this lib is about. Would rather use a boring name like `ReactiveStreamsAdapters` or something like that, also easier to find using autocompletion or searching-for-classname.

[...]

PR #402, opened 2017-11-02
Implementing renaming proposal for the interop with j.u.c.Flow

viktorklang:
As suggested in #401


viktorklang:
@reactive-streams/contributors Please chime in here if you intend to give feedback, so we're not holding off merging this for naught.

viktorklang:
@reactive-streams/contributors Alright, merging!

Issue #284, opened 2015-07-03
Clarification on multiple subscribers needed ..


RuedigerMoeller:
I implemented multiple subscribers using a "wait-for-slowest-subscriber" policy. This means if of 2 subscribers only one sends "credits" via request(N), no message will be sent. So credits = Min(credits of subscribers)

this makes 

`required_mustRequestFromUpstreamForElementsThatHaveBeenRequestedLongAgo()`
and `required_spec104_mustCallOnErrorOnAllItsSubscribersIfItEncountersANonRecoverableError()` fail.

But "wait-for-slowest-subscriber" (with optional timout/forced cancel policy) actually does not violate the spec, isn't it ?

ktoso:
Hi @RuedigerMoeller, thanks a lot for catching this - you're absolutely correct.
These tests assume too much about the publisher - namely that it just signals the elements once demand comes in. I've skimmed the impls now and it seems that amending them won't be too hard - I'll send in a PR soon to fix this.

RuedigerMoeller:
Thanks for the quick response :). Kudos for the test suite, it discovered an awful lot of issues in my initial implementation ..
Hoping to provide a hp multiple-transport reactive-streams remoting plug soon :)

ktoso:
Addressed one part of the issue in this PR: https://github.com/reactive-streams/reactive-streams-jvm/pull/287
Will try to make time and figure out `required_mustRequestFromUpstreamForElementsThatHaveBeenRequestedLongAgo` soon.

RuedigerMoeller:
Hi, afair `required_mustRequestFromUpstreamForElementsThatHaveBeenRequestedLongAgo ` had the same problem in the test prologue.
Will unignore the testcase and retry when out of office ;)

viktorklang:
@ktoso Was this fixed and merged?

viktorklang:
@ktoso Sorry, I see the PR…

ktoso:
Yeap, it's that one which I need to get back to: https://github.com/reactive-streams/reactive-streams-jvm/pull/287

viktorklang:
@RuedigerMoeller Please see #414. Will be released in 1.0.2 which is due before end-of-year!

PR #414, opened 2017-11-23
Allow request-coordinating Processor implementations to pass the TCK

akarnokd:
This PR fixes the TCK to support request-coordinating `Processor` implementations. These kinds of `Processor`s may either 

- coordinate the request amounts of their `Subscriber`s and request only from upstream when all `Subscriber`s requested something; or
- coordinate emissions, requesting a bounded amount upfront from the upstream and then emitting only when all `Subscriber`s have requested something.

From the downstream `Subscriber`s' perspective, both manifest as lack of emissions.

The `IdentityProcessorVerification.doesCoordinatedEmission()` added will affect

- `required_spec104_mustCallOnErrorOnAllItsSubscribersIfItEncountersANonRecoverableError`
- `required_mustRequestFromUpstreamForElementsThatHaveBeenRequestedLongAgo`

By adjusting the test request pattern to request on both `Subscriber`s before asserting on events received.

The `LockstepProcessorTest` contains the `LockstepProcessor` implementation which coordinates emissions when there are multiple `Subscriber`s.

Replaces #287.
Original issue #284.

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
@@ -464,13 +464,13 @@ public abstract class IdentityProcessorVerification<T> extends WithHelperPublish
 
             expectRequest(env.defaultTimeoutMillis(),
                     "If the Processor coordinates requests/emissions when having multiple Subscribers"
-                    + " at once, please override doesCoordinatedEmission() in this "
+                    + " at once, please override doesCoordinatedEmission() to return true in this "
                     + "IdentityProcessorVerification to allow this test to pass.");
 
             final T x = sendNextTFromUpstream();
             expectNextElement(sub1, x,
                     "If the Processor coordinates requests/emissions when having multiple Subscribers"
-                            + " at once, please override doesCoordinatedEmission() in this "
+                            + " at once, please override doesCoordinatedEmission() to return true in this "
                             + "IdentityProcessorVerification to allow this test to pass.");
 
             sub1.request(1);
@@ -757,7 +757,7 @@ public abstract class IdentityProcessorVerification<T> extends WithHelperPublish
             final T z = sendNextTFromUpstream();
             expectNextElement(sub1, z,
                     "If the Processor coordinates requests/emissions when having multiple Subscribers"
-                            + " at once, please override doesCoordinatedEmission() in this "
+                            + " at once, please override doesCoordinatedEmission() to return true in this "
                             + "IdentityProcessorVerification to allow this test to pass.");
             sub2.expectNone(); // since sub2 hasn't requested anything yet
 
--- a/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
@@ -424,7 +424,7 @@ public abstract class IdentityProcessorVerification<T> extends WithHelperPublish
    * <p>
    * The test is not executed if {@link IdentityProcessorVerification#maxSupportedSubscribers()} is less than 2.
    * <p>
-   * If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * If this test fails, the following could be checked within the {@code Processor} implementation:
    * <ul>
    * <li>The {@code TestEnvironment} has large enough timeout specified in case the {@code Processor} has some time-delay behavior.</li>
    * <li>The {@code Processor} is able to fulfill requests of its {@code Subscriber}s independently of each other's requests or
@@ -709,7 +709,7 @@ public abstract class IdentityProcessorVerification<T> extends WithHelperPublish
    * <p>
    * The test is not executed if {@link IdentityProcessorVerification#maxSupportedSubscribers()} is less than 2.
    * <p>
-   * If this test fails, the following could be checked within the {@code Publisher} implementation:
+   * If this test fails, the following could be checked within the {@code Processor} implementation:
    * <ul>
    * <li>The {@code TestEnvironment} has large enough timeout specified in case the {@code Processor} has some time-delay behavior.</li>
    * <li>The {@code Processor} is able to fulfill requests of its {@code Subscriber}s independently of each other's requests or
--- a/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
@@ -189,6 +189,16 @@ public abstract class IdentityProcessorVerification<T> extends WithHelperPublish
       return Long.MAX_VALUE;
   }
 
+  /**
+   * Override this method and return {@code true} if the {@link Processor} returned by the
+   * {@link #createIdentityProcessor(int)} coordinates its {@link Subscriber}s
+   * request amounts and only delivers onNext signals if all Subscribers have
+   * indicated (via their Subscription#request(long)) they are ready to receive elements.
+   */
+  public boolean doesCoordinatedEmission() {
+    return false;
+  }
+
   ////////////////////// TEST ENV CLEANUP /////////////////////////////////////
 
   @BeforeMethod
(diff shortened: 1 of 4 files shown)


tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
```diff
@@ -415,17 +425,33 @@ public TestSetup apply(Long aLong) throws Throwable {
           final ManualSubscriberWithErrorCollection<T> sub2 = new ManualSubscriberWithErrorCollection<T>(env);
           env.subscribe(processor, sub2);
 
-          sub1.request(1);
-          expectRequest();
-          final T x = sendNextTFromUpstream();
-          expectNextElement(sub1, x);
-          sub1.request(1);
+          final Exception ex = new RuntimeException("Test exception");
```
viktorklang: Since most devs probably won't ever notice this method, or know that they should enable/disable it—what do you think about including a hint about it in case the tests which use it fail?

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
```diff
@@ -189,6 +189,16 @@ public long maxSupportedSubscribers() {
       return Long.MAX_VALUE;
   }
 
+  /**
+   * Override this method and return {@code true} if the {@link Processor} returned by the
+   * {@link #createIdentityProcessor(int)} coordinates its {@link Subscriber}s
+   * request amounts and only delivers onNext signals if all Subscribers have
+   * indicated (via their Subscription#request(long)) they are ready to receive elements.
+   */
+  public boolean doesCoordinatedEmission() {
```
viktorklang: Since most devs probably won't ever notice this method, or know that they should enable/disable it—what do you think about including a hint about it in case the tests which use it fail?

(Sorry, this comment was on the wrong line before)

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
```diff
@@ -415,17 +444,33 @@ public TestSetup apply(Long aLong) throws Throwable {
...
-          sendError(ex);
+            final T x = sendNextTFromUpstream();
+
+            expectNextElement(sub1, x);
+            expectNextElement(sub2, x);
+
+            sub1.request(1);
+            sub2.request(1);
+
+            sendError(ex);
```
viktorklang: Perhaps put `sendError(ex);` after the branch (since it is the last statement in each "leg" of the branch.

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
```diff
@@ -673,15 +736,29 @@ public TestSetup apply(Long subscribers) throws Throwable {
...
-          expectNextElement(sub2, z);
+            sub2.request(1);
 
-          if (totalRequests == 3) {
-            expectRequest();
+            final T z = sendNextTFromUpstream();
+            expectNextElement(sub1, z);
+            expectNextElement(sub2, z);
+
+            if (totalRequests == 3) {
```
viktorklang: Since this is the same behavior on each side of the branch, perhaps put it after the branch?

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
```diff
@@ -401,8 +411,27 @@ public void required_spec317_mustNotSignalOnErrorWhenPendingAboveLongMaxValue()
...
+   * <p>
+   * If the {@code Processor} requests and/or emits items only when all of its {@code Subscriber}s have requested,
+   * override {@link #doesCoordinatedEmission()} and return {@code true} to indicate this property.
+   * <p>
+   * <b>Verifies rule:</b> <a href='https://github.com/reactive-streams/reactive-streams-jvm#1.4'>1.4</a> with multiple
+   * {@code Subscriber}s.
+   * <p>
+   * The test is not executed if {@link IdentityProcessorVerification#maxSupportedSubscribers()} is less than 2.
+   * <p>
+   * If this test fails, the following could be checked within the {@code Publisher} implementation:
```
viktorklang: If we wanted to include a hint about `doesCoordinatedEmission` either when running the test case, or when the testcase fails, what would you propose?

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
```diff
@@ -415,17 +444,31 @@ public TestSetup apply(Long aLong) throws Throwable {
...
+            sub2.request(1);
 
-          final Exception ex = new RuntimeException("Test exception");
+            final T x = sendNextTFromUpstream();
+
+            expectNextElement(sub1, x);
+            expectNextElement(sub2, x);
+
+            sub1.request(1);
+            sub2.request(1);
```
viktorklang: Shouldn't `expectReuest()` work after these two?

viktorklang:
@akarnokd Thanks for taking a stab at this issue!

@ktoso, @reactive-streams/contributors Anyone else have feedback on this?

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
akarnokd: Yes, it should. I'll update.

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
akarnokd: We could introduce an overload to `expectNextElement` that takes an error message that hints about that method.

viktorklang: :+1:

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
```diff
@@ -673,13 +745,25 @@ public TestSetup apply(Long subscribers) throws Throwable {
...
-          sub2.request(1);
-          expectNextElement(sub2, z);
+            final T z = sendNextTFromUpstream();
+            expectNextElement(sub1, z);
+            expectNextElement(sub2, z);
+          } else {
+            final T z = sendNextTFromUpstream();
+            expectNextElement(sub1, z,
+                    "If the Processor coordinates requests/emissions when having multiple Subscribers"
+                            + " at once, please override doesCoordinatedEmission() in this "
```
viktorklang: "please override doesCoordinatedEmission() to return `true` in this"

akarnokd: Updated.

ktoso:
I do want to review this, please don't merge yet. 
Esp. I'm worried about the use of coordinated omission -- I don't think this is an instance of that, and the term is used a bit misleadingly. Coordinated omission is about a measurement problem where the measuring-party is not measuring the real latencies due to backpressure from stressed system, I don't think that a Processor that handles many subscribers sis the same thing.

I want to review the thing in depth to be able to reply more precisely though

viktorklang:
@ktoso Are you conflating "coordinated omission" with "coordinated emission"?

In any case, perhaps there's a better term, I'm all for improvements when it comes to names :)

ktoso:
Not conflating, but misreading what was in the PR it seems 😉

viktorklang:
@ktoso No worries! :-)

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
ktoso: That sounds very good! (the hint override)

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
```diff
@@ -748,6 +832,13 @@ public void expectNextElement(ManualSubscriber<T> sub, T expected) throws Interr
       }
     }
 
+    public void expectNextElement(ManualSubscriber<T> sub, T expected, String errorMessageAddendum) throws InterruptedException {
+      final T elem = sub.nextElement(String.format("timeout while awaiting %s. %s", expected, errorMessageAddendum));
+      if (!elem.equals(expected)) {
+        env.flop(String.format("Received `onNext(%s)` on downstream but expected `onNext(%s)`", elem, expected));
+      }
+    }
```
ktoso: Nice 👍

viktorklang:
@ktoso @reactive-streams/contributors Ready for merge?

tck/src/main/java/org/reactivestreams/tck/IdentityProcessorVerification.java
akarnokd: I can add some description to it in a couple of hours.

akarnokd: Done.

viktorklang:
VERY nice, @akarnokd! Merging!

Issue #416, opened 2017-11-29
TCK Improvement: Throw meaningful error in white box subscriber verification when Probe.registerOnSubscribe isn't called


jroper:
Currently, none of the the `SubscriberWhiteboxVerification` tests check whether `WhiteboxSubscriberProbe.registerOnSubscribe` is invoked before it starts trying to use the puppet supplied by that invocation. So, if it wasn't invoked, it just throws an NPE:

```
java.lang.NullPointerException
	at org.reactivestreams.tck.SubscriberWhiteboxVerification$1.run(SubscriberWhiteboxVerification.java:83)
	at org.reactivestreams.tck.SubscriberWhiteboxVerification.subscriberTest(SubscriberWhiteboxVerification.java:497)
	at org.reactivestreams.tck.SubscriberWhiteboxVerification.required_exerciseWhiteboxHappyPath(SubscriberWhiteboxVerification.java:80)
```

It would be nice if this was checked, and a meaningful error message (saying that `registerOnSubscribe` hasn't been invoked) was thrown.

ktoso:
Sounds like a good improvement, are you giving it a shot or should I try to next week (won't be able this week)?

jroper:
I'll take a look today.

viktorklang:
@jroper If the promise hasn't been completed at the time of trying to use it, it *should* yield a test failure:

    public T value() {
      if (isCompleted()) {
        return _value;
      } else {
        env.flop("Cannot access promise value before completion");
        return null;
      }
    }

Perhaps that failure is hidden by an NPE? In which case, we might want to both flop AND throw?

viktorklang:
@jroper are these tests running without `runDefaultInit`? https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java#L541

viktorklang:
@jroper I'm planning on issuing the 1.0.2-RC1 within 24h and would love to include this improvement as well. Will you have time to take a stab at it, or can provide a "failing test" someone could use to verify if it gets fixed? Thank you!

viktorklang:
@jroper 6h to go

akarnokd:
@viktorklang Do you want the fix for this included in RC1? If so, I can have a look and post a fix today.

viktorklang:
[...]

PR #418, opened 2017-12-04
TCK: Fail early if registerOnSubscribe is not called

akarnokd:
This PR makes sure the TCK checks if the `WhiteboxSubscriberProbe.registerOnSubscribe()` was called before any test body tries to use the (unavailable) puppet, crashing with `NullPointerException` and hiding the fact the method call was missing.

Fixes: #416


tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
```diff
@@ -539,6 +539,7 @@ public WhiteboxTestStage(TestEnvironment env, boolean runDefaultInit) throws Int
         probe = new WhiteboxSubscriberProbe<T>(env, subscriber);
         subscribe(createSubscriber(probe));
         probe.puppet.expectCompletion(env.defaultTimeoutMillis(), String.format("Subscriber %s did not `registerOnSubscribe`", sub()));
+        env.verifyNoAsyncErrorsNoDelay();
```
viktorklang: Nicely done, @akarnokd! :+1:

viktorklang:
@reactive-streams/contributors Alright, merging this TCK change as it doesn't impact any RS semantics.

PR #421, opened 2017-12-18
Releasing version 1.0.2

viktorklang:
For 1.0.2. Same as v1.0.2 tag


ktoso:
Artifacts are on their way to maven central.

PR #423, opened 2018-01-02
Ensure TCK passes with empty publisher that synchronously completes

### LINKED ISSUE #422: PublisherVerification.optional_spec105_emptyStreamMustTerminateBySignallingOnComplete fails if the publisher completes synchronously
Consider the following empty publisher:

```java
class EmptyPublisher implements Publisher<Void> {
  public void subscribe(Subscriber<? super Void> subscriber) {
    Objects.requireNonNull(subscriber);
    subscriber.onSubscribe(new Subscription() { ... });
    subscriber.onComplete();
  }
}
```

I believe this is a valid publisher according to the spec, there's nothing that says that `onComplete` can't be synchronously invoked from `subscribe`, it just has to be invoked after `onSusbscribe` (it is). 

However `PublisherVerification.optional_spec105_emptyStreamMustTerminateBySignallingOnComplete`, a test for empty publishers which should pass on the above publisher, fails with `java.lang.AssertionError: Subscriber::onComplete() called before Subscriber::onSubscribe`.

The problem seems to be that `ManualSubscriberWithSubscriptionSupport` requires `expectCompletion` to be invoked after `onSubscribe` and before `onComplete` is invoked. But `expectCompletion` isn't invoked until after `subscribe` returns, hence synchronous invocation of `onComplete` after `onSusbcribe` in  the `subscribe` method will always fail.

jroper:
Fixes #422

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
+++ b/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
@@ -24,6 +24,7 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import static org.testng.Assert.assertTrue;
 import static org.testng.Assert.fail;
@@ -570,6 +571,8 @@ public class TestEnvironment {
 
   public static class ManualSubscriberWithSubscriptionSupport<T> extends ManualSubscriber<T> {
 
+    private final AtomicBoolean onSubscribeCalled = new AtomicBoolean();
+
     public ManualSubscriberWithSubscriptionSupport(TestEnvironment env) {
       super(env);
     }
@@ -587,7 +590,7 @@ public class TestEnvironment {
     @Override
     public void onComplete() {
       env.debug(this + "::onComplete()");
-      if (subscription.isCompleted()) {
+      if (onSubscribeCalled.get()) {
         super.onComplete();
       } else {
         env.flop("Subscriber::onComplete() called before Subscriber::onSubscribe");
@@ -599,6 +602,7 @@ public class TestEnvironment {
       env.debug(String.format("%s::onSubscribe(%s)", this, s));
       if (!subscription.isCompleted()) {
         subscription.complete(s);
+        onSubscribeCalled.set(true);
       } else {
         env.flop("Subscriber::onSubscribe called on an already-subscribed Subscriber");
       }
@@ -607,7 +611,7 @@ public class TestEnvironment {
     @Override
     public void onError(Throwable cause) {
       env.debug(String.format("%s::onError(%s)", this, cause));
-      if (subscription.isCompleted()) {
+      if (onSubscribeCalled.get()) {
         super.onError(cause);
       } else {
         env.flop(cause, String.format("Subscriber::onError(%s) called before Subscriber::onSubscribe", cause));
--- a/tck/src/test/java/org/reactivestreams/tck/flow/support/TCKVerificationSupport.java
+++ b/tck/src/test/java/org/reactivestreams/tck/flow/support/TCKVerificationSupport.java
@@ -81,6 +81,18 @@ public class TCKVerificationSupport {
     throw new RuntimeException("Expected TCK to SKIP this test, instead if PASSed!");
   }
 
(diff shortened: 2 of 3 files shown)


tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -599,6 +602,7 @@ public void onSubscribe(Subscription s) {
       env.debug(String.format("%s::onSubscribe(%s)", this, s));
       if (!subscription.isCompleted()) {
         subscription.complete(s);
```
akarnokd: There is a [completeImmediately()](https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java#L910) method on the promise.

akarnokd: Btw. the plain `complete()` behavior is odd. I'd think once called, `isCompleted()` should report true but the implementation queues the object.

jroper: I agree - but I'm not familiar enough with the TCK to change it. Though I can understand why the event is queued, so that the order of events can be verified and duplicates can be detected etc.

jroper:
I thought about changing `onNext`, however, the manual subscriber will never invoke `Subscription.request` before invoking `expectCompletion`, so I think that logic is still valid, even if the error message is a little odd. My worry with changing it is that there might be some tests that depend on the current behaviour.

tck/src/test/java/org/reactivestreams/tck/flow/support/TCKVerificationSupport.java
```diff
@@ -81,6 +81,18 @@ public void requireTestSkip(ThrowingRunnable run, String msgPart) {
     throw new RuntimeException("Expected TCK to SKIP this test, instead if PASSed!");
   }
 
+  public void requireOptionalTestPass(ThrowingRunnable run) {
+    try {
+      run.run();
+    } catch (SkipException skip) {
+      throw new RuntimeException("Expected TCK to PASS this test, instead it was SKIPPED", skip.getCause());
+    } catch (Throwable throwable) {
+      throw new RuntimeException(
```
viktorklang: Repackage Errors as Exceptions?

jroper: Just being consistent with `requireTestFailure` and `requireTestSkip`.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
ktoso: I'll investigate this

tck/src/test/java/org/reactivestreams/tck/flow/support/TCKVerificationSupport.java
ktoso: Does not really matter much as those will end up failing the test in the suite and one may always look into the logs. It's not like it's a production app where we have to watch out to not accidentally continue after an OOM or something... Fine as is I think

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
ktoso: Thanks for noticing, I think we can improve the impl there. I'll work on it today after done with work stuff.
I'll PR into this PR then :)

ktoso:
I submitted an impl improvement of the Promise which makes the test pass the new requirement: 
https://github.com/jroper/reactive-streams-jvm/pull/1

I submitted the change as PR into this PR, so if you accept it over there @jroper we could merge all those commits rebased here together I guess.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
ktoso: I reimplemented the Promise which makes this workwithout the additional atomic.
Please see here: https://github.com/jroper/reactive-streams-jvm/pull/1

viktorklang:
@jroper @ktoso Hey all, I just wanted to know what the current status is here.

ktoso:
I would like @jroper to merge the fix I proposed to this PR, it's been waiting ;-)

ktoso:
PR was merged, thanks @jroper ! Seems good to go from my side then

viktorklang:
thanks @jroper!

Issue #427, opened 2018-03-12
TCK should stress subscriber rule 2.5 in concurrent setting


ktoso:
By now I have seen a number of talks where people present Reactive Streams / java.util.concurrent.Flow, and end up implementing their own non-compliant implementations. This, by itself, is not very good as it creates the false impression of trivialness of implementing these interfaces.

One common mistake seems to be forgetting about concurrency in the `Subscriber#onSubscribe` call, e.g. by naively implementing it like this: 

```java
            new Subscriber() {

              Subscription sub = null; // on purpose no volatile or any external synchronization

              @Override
              public void onSubscribe(Subscription s) {
               // this is racy, and we may accept (illegally?) multiple subscriptions
                if (this.sub == null) {
                  this.sub = s;
                  s.request(1);

// ...
```

This is racy, and may accept multiple subscriptions if called concurrently, yet almost all talks I've seen that present j.u.c.Flow do this mistake. While improving the TCK may not help those who do not use it, adding a test that exercises this case concurrently would be a good addition to the TCK.

I will submit a PR now.

ktoso:
Reminder of the rule:

> 2.5 A Subscriber MUST call Subscription.cancel() on the given Subscription after an onSubscribe signal if it already has an active Subscription.

akarnokd:
`onSubscribe` must be called sequentially and a `Publisher` should not call it more than once. We can only discourage reusing the same `Subscriber` with multiple `Publisher`s which may actually lead to such `onSubscribe` calls. I don't think we should actively defend against possible concurrent invocations of `onSubscribe`. 

Defending against synchronous multiple invocations may be reasonable, but unfortunately so far, it only created a lot of partial code coverage results. This is forcing me to actively test out such defenses, even though it was practically never a case when an RxJava operator was buggy in this regard - a missing call to `onSubscribe` was way more common in the early days.

The second problem is more like with `s.request(1)` followed by non-synchronized state manipulations which may end up interfering with a concurrent `onNext` invocation triggered by the request call. Deferring a request until `onSubscribe()` returns adds some overload and we generally don't do it in RxJava unless a `Subscriber` is an arbitrary Reactive Streams implementation (instead of our own `FlowableSubscriber`). In fact, this deferred requesting is mandated by the TCK currently as some tests fail reporting concurrent `onSubscribe`-`onNext` invocations.

[...]

PR #428, opened 2018-03-12
=tck #427 TCK should stress subscriber rule 2.5 in concurrent setting

ktoso:
Test implementation for #427 TCK should stress subscriber rule 2.5 in concurrent setting

Discussion so far in the ticket.

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/SubscriberBlackboxVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/SubscriberBlackboxVerification.java
@@ -25,8 +25,11 @@ import org.testng.annotations.BeforeClass;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static org.reactivestreams.tck.SubscriberWhiteboxVerification.BlackboxSubscriberProxy;
 import static org.testng.Assert.assertTrue;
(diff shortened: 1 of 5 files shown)


tck/src/main/java/org/reactivestreams/tck/SubscriberBlackboxVerification.java
```diff
@@ -231,6 +234,77 @@ public String toString() {
     }};
   }
 
+  @Override @Test
+  public void required_spec205_blackbox_mustCallSubscriptionCancelIfItAlreadyHasAnSubscriptionAndReceivesAnotherOnSubscribeSignalConcurrently() throws Throwable {
```
akarnokd: Travis may not have the proper concurrency to verify this. I'm constantly facing the problem testing race conditions in RxJava requiring two dedicated cores running practically at the same time to trigger small windows around CAS instructions.

This could be the reason the build failed: https://travis-ci.org/reactive-streams/reactive-streams-jvm/jobs/352388267#L1483

tck/src/main/java/org/reactivestreams/tck/SubscriberBlackboxVerification.java
```diff
@@ -231,6 +234,77 @@ public String toString() {
...
+            return null;
+          }
+        };
+
+        List<Callable<Void>> invokeOnSubscribeConcurrentlyTasks = new ArrayList<Callable<Void>>(tasksToSubmit);
+        for (int i = 0; i < tasksToSubmit; i++) {
+          invokeOnSubscribeConcurrentlyTasks.add(invokeOnSubscribe);
+        }
+
+        List<Future<Void>> futures = Executors.newFixedThreadPool(16).invokeAll(invokeOnSubscribeConcurrentlyTasks);
```
akarnokd: Shouldn't this Executor shut down when no longer needed? Also having more threads than cores may actually reduce the chance of hitting the desired concurrency effect because of the increased context switching among threads.

ktoso: Yes it should; first let's agree if this PR makes sense or not -- as we started in the ticket, if not, no reason to spend time polishing it ;-)

ktoso:
Closing, no need to guard against this -- it rides on guarantees made by Publisher in 1.3

Issue #424, opened 2018-02-21
FlowAdapters splits the org.reactivestreams package


jroper:
It is impossible to consume the reactive streams flow adapters in JDK9 as a module since it splits the `org.reactivestreams` package across itself and the `org.reactivestreams` module (which it depends on). Here's an example error message from javac when you try to use it:

```
error: module org.reactivestreams.tck reads package org.reactivestreams from both org.reactivestreams.flowadapters and org.reactivestreams
```

JDK9 does not allow packages split between modules. `FlowAdapters` must be moved to its own package if its to be in its own artifact.

viktorklang:
But moving something to a new pqckahe is not backwards compatible.

Is there any tooling that we should have been using to verify this problem?

What are our options here?

-- 
Cheers,
√

viktorklang:
I guess we can work around it by having the latest version of the flow-adapters jar but the non-automatic-module-name  version of the API-jar? Or?

viktorklang:
@reactive-streams/contributors Any thoughts or suggestions here?

akarnokd:
No clue. I have no experience with Java 9 modules but fixing this sounds like we have to break the structure of library code already released.

jroper:
We could just include the adapters in the main reactive streams jar. It won't cause any problems for users of JDK8 and earlier, since they just won't use them.

We could also use MR jars to make `org.reactivestreams.*` extend `juc.Flow.*`, however the JDK maintainers have explicitly stated that that's not an intended use case for MR jars, so we'd be on shaky ground, particularly when it comes to whether other tooling (IDEs, compilers for other JVM languages) supported it.

ktoso:
> We could just include the adapters in the main reactive streams jar. It won't cause any problems for users of JDK8 and earlier, since they just won't use them.

Yes, this works; 

Akka has been doing this since JDK9 was released: https://github.com/akka/akka/commit/349a5f749ae401ca815482233af3b8d4d2bf67cd The existence of such JDK9-refering classes in a JAR does not cause problems for JDK8 users.

The classes can remain "where they are" package-wise then, so the change is source as well as binary compatible.

It also means deprecating the adapter artifact as it would not be intended to be used. It may happen that someone ends up with both artifacts on the classpath... "new" RS that contains adapters, and "old RS adapters", however since the classes are to be exactly the same that's managable risk... and should not really happen one would hope (though stating the fact that that's the risk/tradeoff we pick when we do this)

[...]

PR #430, opened 2018-05-02
Moving the FlowAdapters into the core jar as a solution to #424

viktorklang:
Solves #424 by moving the FlowAdapters into the main jar and retires the flow-adapters artifact.

@reactive-streams/contributors Thoughts?


viktorklang:
@akarnokd Everything will be compiled with source and target 1.6, but FlowAdapters will only be linked/used in the build if compiled against jdk9+ (conditional sources).

viktorklang:
@reactive-streams/contributors Anyone?

ktoso:
Sanity check the:

> Everything will be compiled with source and target 1.6, but FlowAdapters will only be linked in the build if compiled against jdk9+

as I'm missing where the magic that does this in our build is here. If it indeed does so, this is ready, except the minor docs addition requested above.

ktoso:
It *has to* be done such that those are java 6 format.
We do so in Akka where most classes are JDK8 yet the Flow.* using ones are JDK9 compiled; the same *has to* be done here with regards to Java 6; which is why I'm asking if we confirmed this actually is happening now.

ktoso:
ping @viktorklang can you confirm the result is the expected one?

viktorklang:
Thanks @ktoso—I didn’t see the previous messages due to a broken filter in
GMail.

👍
-- 
Cheers,
√

viktorklang:
@reactive-streams/contributors Merging!

akarnokd:
Some users on Android + ProGuard have run into some issues due to the mixed RS jar file: https://github.com/ReactiveX/RxJava/issues/6670. This would affect any library depending on RS 1.0.3+ under similar circumstances.

Would the README.md of this repo a good place to document this case and the workarounds?

Issue #410, opened 2017-11-09
Compliance with Rule 2.5


pavelrappo:
> A `Subscriber` MUST call `Subscription.cancel()` on the given `Subscription` after an `onSubscribe` signal if it already has an active `Subscription`.

I think it's worth pointing out in examples (and may be even producing a TCK test), that subscribers must use appropriate synchronization in `onSubscribe`. Even though there is rule 1.3:

> `onSubscribe`, `onNext`, `onError` and `onComplete` signaled to a `Subscriber` MUST be signaled in a thread-safe manner—and if performed by multiple threads—use external synchronization.

multiple `Publishers` may not coordinate their actions for a subscriber, thus `onSubscribe` may be called concurrently even though each of the publishers call it in a thread-safe manner.

For example, synchronous subscriber could use something like this:

```java
            private final AtomicBoolean hasSubscription = new AtomicBoolean();
            private Subscription subscription;
            
            @Override 
            public void onSubscribe(Subscription s) {
                Objects.requireNonNull(s);
                if (!hasSubscription.compareAndSet(false, true)) {
                    s.cancel();
                } else {
                    subscription = s;
                    // ...
                }
            }
```

ktoso:
Good idea, indeed a Subscriber must be defensive in its onSubscribe implementation, even though for each of the individual pairs we're "safe" due to the 1.3 rule that we impose onto Publishers.

Our test for 2.5 covers this but in a sequential simple setting, we can add one to issue the onSubscribes concurrently indeed.

pavelrappo:
@ktoso I also found a case (even though an extremely unlikely one) which permits an incoherent calls to a `Subscriber`'s methods should a `Publisher` attempt to subscribe this subscriber after the subscriber has been already subscribed to another publisher. Even if calls to this subscriber's methods happen in a thread-safe manner across all publishers.

Consider that even if a `Subscriber` calls `Subscription.cancel` in `onSubscribe` after identifying it already has an active `Subscription`, the `Publisher` can still deliver `onComplete` or `onError` due to a possible propagation delay ([rule 1.8](https://github.com/reactive-streams/reactive-streams-jvm#1.8)). Worst-case scenario -- an unrelated call to `onComplete` or `onError` potentially leading to  an arbitrary size of mess in processing.

Unfortunately in this case not much can be done, but hoping `Subscription.cancel` is responsive enough to terminate the subscription immediately.

viktorklang:
[...]

PR #437, opened 2018-08-20
Adds clarification to the intents section of 2.5 that the condition m…

viktorklang:
…ight not always be detectable.

As discussed in #410

@reactive-streams/contributors Please weigh in here. Another option is to weaken 2.5 to be a SHOULD rule, or taking a harder stance on 2.5 and encourage the use of atomics for setting the active Subscription.


DougLea:
If this is weakened, something should be said about what happens if subscribers do not do this: that they may get either no or multiple signals for subsequent elements? (We do not want to just say "undefined behavior".)

viktorklang:
@DougLea How about something like: "Failure to enforce this rule can lead to hard to diagnose errors due to erronous signalling patterns if a Subscriber instance is reused."

viktorklang:
@DougLea This issue is akin to being able to detect ConcurrentModification in collections for instance.

DougLea:
We could say that failure to conform may lead to violations of Publisher rule 1 (The total number of onNext´s signalled by a Publisher to a Subscriber MUST be less than or equal to the total number of elements requested by that Subscriber´s Subscription at all times.) And other rules too, but mainly this one.

And then maybe note in rationale that such violations lead to hard -to-diagnose bugs.

viktorklang:
@DougLea That makes a lot of sense. I'll amend this PR.

viktorklang:
@DougLea & @reactive-streams/contributors I've updated the text, let me know what you think.

DougLea:
OK, but you could simplify:
"prevent that two, or more, separate Publishers from thinking that they can interact" =>
"prevent two or more Publishers from trying to interact"

viktorklang:
@DougLea Incorporated your suggestion!

viktorklang:
@reactive-streams/contributors Ping. Will merge at the end of the week unless someone objects :)

viktorklang:
@reactive-streams/contributors Merging without objection!

PR #441, opened 2018-11-18
Wait for demand before signaling onNext

### LINKED ISSUE #277: required_spec208_mustBePreparedToReceiveOnNextSignalsAfterHavingCalledSubscriptionCancel does not wait for request before invoking onNext
``` java
  @Override @Test
  public void required_spec208_mustBePreparedToReceiveOnNextSignalsAfterHavingCalledSubscriptionCancel() throws Throwable {
    subscriberTest(new TestStageTestRun() {
      @Override
      public void run(WhiteboxTestStage stage) throws InterruptedException {
        stage.puppet().triggerRequest(1);
        stage.puppet().signalCancel();
        stage.signalNext();

        stage.puppet().triggerRequest(1);
        stage.puppet().triggerRequest(1);

        stage.verifyNoAsyncErrors();
      }
    });
  }
```

Note, it triggers the request, but then does not wait for the request to arrive at the subscription before signalling next.  This violates the spec, and is causing my tests to fail, since my subscriber setup and request triggering is asynchronous, so my subscriber under test is not necessarily ready to receive elements before requesting them.

jroper:
Fixes #277.

Fixes two whitebox subscriber tests where they weren't waiting for demand before signaling onNext.

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
@@ -267,6 +267,7 @@ public abstract class SubscriberWhiteboxVerification<T> extends WithHelperPublis
       public void run(WhiteboxTestStage stage) throws InterruptedException {
         stage.puppet().triggerRequest(1);
         stage.puppet().signalCancel();
+        stage.expectRequest();
         stage.signalNext();
 
         stage.puppet().triggerRequest(1);
@@ -437,7 +438,12 @@ public abstract class SubscriberWhiteboxVerification<T> extends WithHelperPublis
       @Override
       public void run(WhiteboxTestStage stage) throws InterruptedException {
         stage.puppet().triggerRequest(2);
+        long requestedElements = stage.expectRequest();
         stage.probe.expectNext(stage.signalNext());
+        // Some subscribers may only request one element at a time.
+        if (requestedElements < 2) {
+          stage.expectRequest();
+        }
         stage.probe.expectNext(stage.signalNext());
 
         stage.probe.expectNone();
--- a/tck/src/test/java/org/reactivestreams/tck/SubscriberWhiteboxVerificationTest.java
+++ b/tck/src/test/java/org/reactivestreams/tck/SubscriberWhiteboxVerificationTest.java
@@ -23,7 +23,10 @@ import org.testng.annotations.Test;
 
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * Validates that the TCK's {@link SubscriberWhiteboxVerification} fails with nice human readable errors.
@@ -31,8 +34,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
  */
 public class SubscriberWhiteboxVerificationTest extends TCKVerificationSupport {
 
-  private ExecutorService ex;
-  @BeforeClass void before() { ex = Executors.newFixedThreadPool(4); }
+  private ScheduledExecutorService ex;
+  @BeforeClass void before() { ex = Executors.newScheduledThreadPool(4); }
   @AfterClass void after() { if (ex != null) ex.shutdown(); }
 
   @Test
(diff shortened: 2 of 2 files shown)


tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
```diff
@@ -267,6 +267,7 @@ public void required_spec208_mustBePreparedToReceiveOnNextSignalsAfterHavingCall
       public void run(WhiteboxTestStage stage) throws InterruptedException {
         stage.puppet().triggerRequest(1);
         stage.puppet().signalCancel();
+        stage.expectRequest();
```
akarnokd: Should be there an unit test that failed before this change?

viktorklang: Thinking about this some more, shouldn't the `expectRequest` be before the `signalCancel`?

akarnokd: Also could it be possible, and legal, that a request followed by a cancel is coalesced into just a cancel?

jroper: > Should be there an unit test that failed before this change?

I'll add one.

> Thinking about this some more, shouldn't the `expectRequest` be before the `signalCancel`?:

I don't think it matters, in fact the reason I put it in this order is that I think this order makes it more likely that the test will test what it's intending to test. Triggering a request, waiting for that request to come, then signalling cancel, then immediately sending the next message, means that there's a high chance this will pass due to signal not being handled yet. However, waiting for the request after signalling cancel increases the likelihood that the cancel signal has been handled before we emit the next element, thereby increasing the likelihood that the test will actually test what it says its testing (that elements emitted after a cancel signal is handled won't cause an error).

> Also could it be possible, and legal, that a request followed by a cancel is coalesced into just a cancel?

On a subscription, yes, that's legal and possible, but this isn't a subscription, this is the TCK puppet, and the TCK defines how implementations of the puppet are are to behave. It requires that the request signal does result in a request for demand to the subscription, regardless of what signal comes immediately after. And when these signals get forwarded onto the subscription, we know the subscription won't coalesce the signals because that is a subscription that the TCK has implemented itself in `ManualPublisher`, and it doesn't coalesce signals, it captures them all for assertions to be run later, assertions like `expectRequest`.

viktorklang: >However, waiting for the request after signalling cancel increases the likelihood that the cancel signal has been handled before we emit the next element, thereby increasing the likelihood that the test will actually test what it says its testing (that elements emitted after a cancel signal is handled won't cause an error).

Since it is, as you say, not a full solution, may I suggest that it makes sense to add that as a comment, since it is non-apparent as to its intent?

viktorklang:
Nice work, @jroper!

viktorklang:
@reactive-streams/contributors Merging this without objections.

viktorklang:
Thanks @jroper! 👍

tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
jrudolph: > On a subscription, yes, that's legal and possible, but this isn't a subscription, this is the TCK puppet, and the TCK defines how implementations of the puppet are are to behave. It requires that the request signal does result in a request for demand to the subscription, regardless of what signal comes immediately after. And when these signals get forwarded onto the subscription, we know the subscription won't coalesce the signals because that is a subscription that the TCK has implemented itself in `ManualPublisher`, and it doesn't coalesce signals, it captures them all for assertions to be run later, assertions like `expectRequest`.

These tests are also used in the `IdentityProcessorVerification` tests which implements the `SubscriberPuppet` as forwarding the triggers as is to the processor under test. So, IIUC the invariants you state would have to be valid for any processor you want to test. That would mean, that you cannot use the TCK to check processor implementations that do coalescing.

jroper: In which case, perhaps we should change this to `expectRequest` before we `signalCancel`, and then insert a new `expectCancelling` after that, before we `signalNext`, to ensure that the subscriber did call cancel before we sent the next element - that's I think the original intent of the ordering of operations here, and matches the description of the test.

jroper: I've implemented this here: https://github.com/reactive-streams/reactive-streams-jvm/pull/462

Issue #280, opened 2015-06-30
Subscriber whitebox verification tests demand


jroper:
So firstly, there's no javadocs on `SubscriberPuppet`.  Is `triggerRequest` meant to trigger a request synchronously? Due to #277, it currently must be triggered synchronously, otherwise some of the tests violate the spec - but that's probably a problem with those tests.

Another thing not explained is whether exactly `elements` must be triggered.  I suspect many subscribers (including the one I'm implementing, netty) do not allow this level of control over them, since the number of elements requested is controlled by buffers, and it's only possible for the test to signal an event to say it's possible to write more elements, not how many elements.  Modifying the logic such that something external can trigger a specific number of elements would undermine the existing logic there, meaning that the TCK isn't testing the real subscriber.

So, as it happens, not requesting the exact number of elements still works, the TCK makes no assertions about how many elements were actually requested.  But because of this, the TCK itself violates the spec, since it may end up sending more elements than were requested.  Consider a streams implementation that only supports requesting one element at a time (such as iteratees), the TCK triggers a request for 2 elements, the iteratee requests one, and then the TCK sends 2 elements, a violation of the spec.  It's only due to implementation details in each subscriber that this works.

So I'd like to propose that these tests be modified in the following way:

`SubscriberPuppet.triggerRequest(long elements)` be renamed to `SubscriberPuppet.ensureDemand(long elements)`, and that the contract be documented to say that when this method is invoked, the subscriber must eventually request at least `elements` number of elements, but it may request those elements in batches of one or more, and is allowed to wait for the requested elements to be received by the `onNext` method before requesting the next batch.  Then the TCK tests should be modified to, instead of just asserting that there was a request, read the request, and then send the number of elements that were requested, looping if necessary to read more requests if the full amount of demand wasn't requested.

viktorklang:
@jroper Why introduce a new method for this? Or am I missing something crucial?

jroper:
It's not introducing a new method, it's renaming a method to match its intent.  The method may not necessarily trigger a request for that many elements, but it must ensure that there is eventually at least enough demand for that many elements.

viktorklang:
If we can rename it without breaking BC that's completely fine by me, as well as adding better docs for it.
I don't have enough spare cycles ATM so anyone, feel free to propose a PR.

jroper:
It's a Java interface, you can't add an abstract method without renaming.

[...]

PR #442, opened 2018-11-19
Clarified the expected behaviour for triggerRequest

jroper:
This addresses some of the problem raised in #280.

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
@@ -789,8 +789,23 @@ public abstract class SubscriberWhiteboxVerification<T> extends WithHelperPublis
    * make implement these calls.
    */
   public interface SubscriberPuppet {
+
     /**
-     * Trigger {@code request(elements)} on your {@link Subscriber}
+     * Ensure that at least {@code elements} are eventually requested by your {@link Subscriber}, if it hasn't already
+     * requested that many elements.
+     * <p>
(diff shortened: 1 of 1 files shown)


tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
```diff
@@ -789,8 +789,23 @@ public void registerOnSubscribe(SubscriberPuppet p) {
...
+     * triggered to be requested by all the invocations of this method.
+     * <p>
+     * Additionally, subscribers are permitted to delay requesting elements until previous requests for elements have
+     * been fulfilled. For example, a subscriber that only requests one element at a time may fulfill the request made
+     * by this method by requesting one element {@code elements} times, waiting for each element to arrive before the
+     * next request is made.
+     * <p>
+     * Before sending any element to the subscriber, the TCK must wait for the subscriber to request that element, and
+     * must be prepared for the subscriber to only request one element at a time, it is not enough for the TCK to
+     * simply invoke this method before sending elements.
```
viktorklang: Has the following semantics been verified throughout the TCK now? (So that the TCK uses the method appropriately, as described)

viktorklang: @jroper Ping :)

jroper: Yes, I believe it has.

viktorklang:
Thanks @jroper! :+1:

PR #443, opened 2018-11-19
Fixed javadocs for skipStochasticTests

### LINKED ISSUE #278: Incomplete documentation on stochastic tests in TCK

jroper:
Fixes #278.

This was already fixed for `PublisherVerification`, but the fix wasn't done on the same method on `IdentityProcessorVerification`.


viktorklang:
@reactive-streams/contributors I'm merging this documentation fix as is. (uncontroversial)

Issue #440, opened 2018-09-25
Incorrect documentation for subscription.request


tferi-da:
I believe [this line](https://github.com/reactive-streams/reactive-streams-jvm/blob/master/api/src/main/java/org/reactivestreams/Subscription.java#L26) is misleading.
According to rule 3.17 if demand exceeds Long.MAX_VALUE the publisher should keep working, and may become unbounded. There's even a [test](https://github.com/reactive-streams/reactive-streams-jvm/commit/a3005b779344bf8ed6047d73d89ad9b39bb1d1a7) which is checking that demand beyond Long.MAX_VALUE is tolerated without errors.

viktorklang:
Hi @tferi-da,

Thanks for opening this issue!

I'm inclined to agree with your assessment,
what do you think about amending it as such:

     * It can be called however often and whenever needed—but if the outstanding cumulative demand ever becomes Long.MAX_VALUE or more, it may be treated by the {@link Publisher} as "effectively unbounded".

tferi-da:
Yes @viktorklang, I think that's much better. Thank you for responding so quickly.

viktorklang:
@reactive-streams/contributors Wdyt?

viktorklang:
@reactive-streams/contributors Ping.

tferi-da:
Thank you for the fix @viktorklang

PR #444, opened 2018-12-06
Updates the Subscription.request javadoc to be clearer about cumulati…

viktorklang:
…ve demand beyond Long.MAX_VALUE

As discussed in #440


jroper:
I like the new wording, describes it very clearly and simply.

PR #450, opened 2019-04-18
Update gradlew wrapper to 4.10.3

Scottmitch:
Motivation:
The existing gradlew wrapper version doesn't support JDK11.

Modifications:
- Update to 4.10.3 which does support JDK11.

Result:
TCK can be built on JDK11.


viktorklang:
Thanks @Scottmitch!

Issue #451, opened 2019-04-18
TCK: Receptacle#expectError timeout approach


Scottmitch:
`defaultTimeoutMillis` is typically used as a "max time to wait until an expected event occurs". For example this value is used as the argument to a `ArrayBlockingQueue#poll(timeout)` [1][2][3] method call and therefore when things work as expected you rarely wait this full duration. However [Receptacle#expectError](https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java#L1030) does a `Thread.sleep(..)` and then checks if there is something present in the error list (which is a [CopyOnWriteArrayList](https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java#L45)). The value for the sleep is typically derived from [defaultTimeoutMillis](https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java#L521-L526) which leads to unconditional delays. This makes it challenging to use `defaultTimeoutMillis` in tests to specify "maximum amount of time to wait until some event occurs". The impacts are if `defaultTimeoutMillis` is set very small you are more likely to see timeouts (especially on over utilized CI servers), but if is large `defaultTimeoutMillis` the tests take a long time to complete.

[1] https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java#L1019
[2] https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java#L978
[3] https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java#L990

Scottmitch:
PR https://github.com/reactive-streams/reactive-streams-jvm/pull/452 submitted to address this issue.

PR #452, opened 2019-04-18
TCK: Receptacle#expectError timeout approach

Scottmitch:
Motivation:
`defaultTimeoutMillis` is typically used as a "max time to wait until an expected event occurs". For example this value is used as the argument to a `ArrayBlockingQueue#poll(timeout)` [1][2][3] method call and therefore when things work as expected you rarely wait this full duration. However [Receptacle#expectError](https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java#L1030) does a `Thread.sleep(..)` and then checks if there is something present in the error list (which is a [CopyOnWriteArrayList](https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java#L45)). The value for the sleep is typically derived from [defaultTimeoutMillis](https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java#L521-L526) which leads to unconditional delays. This makes it challenging to use `defaultTimeoutMillis` in tests to specify "maximum amount of time to wait until some event occurs". The impacts are if `defaultTimeoutMillis` is set very small you are more likely to see timeouts (especially on over utilized CI servers), but if `defaultTimeoutMillis` the tests take a long time to complete.

[1] https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java#L1019
[2] https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java#L978
[3] https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java#L990

Modifications:
- TestEnvironment now supports an additional poll timeout parameter to be used
  when there is no explicit async event to trigger off of.

Result:
TestEnvironment's defaultTimeoutMillis can be used as "max time until some event
occurs" where defaultPollTimeoutMillis provides the polling interval to check
for an event if there is no explicit async event to trigger off of.
Fixes https://github.com/reactive-streams/reactive-streams-jvm/issues/451

### DIFF
--- a/tck-flow/README.md
+++ b/tck-flow/README.md
@@ -209,32 +209,39 @@ within the TCK which await for something to happen. The other timeout is `publis
 [Rule 3.13](https://github.com/reactive-streams/reactive-streams-jvm#3.13) which defines that `Subscriber` references MUST be dropped
 by the Publisher.
 
-Note that the TCK differenciates between timeouts for "waiting for a signal" (``defaultTimeoutMillis``),
-and "asserting no signals happen during a given amount of time" (``envDefaultNoSignalsTimeoutMillis``).
-While the latter defaults to the prior, it may be useful to tweak them independently when running on continious 
-integration servers (for example, keeping the no-signals timeout significantly lower).
+Note that the TCK differentiates between timeouts for "waiting for a signal"
+(`defaultTimeoutMillis`), and "asserting no signals happen during a given amount of time"
+(`defaultNoSignalsTimeoutMillis`). While the latter defaults to the prior, it may be useful to tweak
+them independently when running on continuous integration servers (for example, keeping the
+no-signals timeout significantly lower). Another configuration option is the "poll timeout" which is
+used whenever an operation has to poll for a `defaultTimeoutMillis` for a signal to appear (most
+often errors), it can then poll and check using the `defaultPollTimeoutMillis`, for the expected
+error, rather than blocking for the full default timeout.
 
-In order to configure these timeouts (for example when running on a slow continious integtation machine), you can either:
+In order to configure these timeouts (for example when running on a slow continuous integration
+machine), you can either:
 
 **Use env variables** to set these timeouts, in which case the you can do:
 
 ```bash
 export DEFAULT_TIMEOUT_MILLIS=100
 export DEFAULT_NO_SIGNALS_TIMEOUT_MILLIS=100
+export DEFAULT_POLL_TIMEOUT_MILLIS=20
 export PUBLISHER_REFERENCE_GC_TIMEOUT_MILLIS=300
 ```
 
 Or **define the timeouts explicitly in code**:
 
 ```java
-public class RangePublisherTest extends FlowPublisherVerification<Integer> {
+public class RangePublisherTest extends PublisherVerification<Integer> {
 
   public static final long DEFAULT_TIMEOUT_MILLIS = 100L;
   public static final long DEFAULT_NO_SIGNALS_TIMEOUT_MILLIS = DEFAULT_TIMEOUT_MILLIS;
-  public static final long PUBLISHER_REFERENCE_CLEANUP_TIMEOUT_MILLIS = 500L;
+  public static final long DEFAULT_POLL_TIMEOUT_MILLIS = 20;
+  public static final long PUBLISHER_REFERENCE_CLEANUP_TIMEOUT_MILLIS = 300L;
 
   public RangePublisherTest() {
-    super(new TestEnvironment(DEFAULT_TIMEOUT_MILLIS, DEFAULT_TIMEOUT_MILLIS), PUBLISHER_REFERENCE_CLEANUP_TIMEOUT_MILLIS);
+    super(new TestEnvironment(DEFAULT_TIMEOUT_MILLIS, DEFAULT_TIMEOUT_MILLIS, DEFAULT_POLL_TIMEOUT_MILLIS), PUBLISHER_REFERENCE_CLEANUP_TIMEOUT_MILLIS);
   }
 
   // ...
--- a/tck-flow/README.md
+++ b/tck-flow/README.md
@@ -233,7 +233,7 @@ export PUBLISHER_REFERENCE_GC_TIMEOUT_MILLIS=300
 Or **define the timeouts explicitly in code**:
 
 ```java
-public class RangePublisherTest extends PublisherVerification<Integer> {
+public class RangePublisherTest extends FlowPublisherVerification<Integer> {
 
   public static final long DEFAULT_TIMEOUT_MILLIS = 100L;
   public static final long DEFAULT_NO_SIGNALS_TIMEOUT_MILLIS = DEFAULT_TIMEOUT_MILLIS;
--- a/tck-flow/README.md
+++ b/tck-flow/README.md
@@ -237,7 +237,7 @@ public class RangePublisherTest extends FlowPublisherVerification<Integer> {
 
   public static final long DEFAULT_TIMEOUT_MILLIS = 100L;
   public static final long DEFAULT_NO_SIGNALS_TIMEOUT_MILLIS = DEFAULT_TIMEOUT_MILLIS;
-  public static final long DEFAULT_POLL_TIMEOUT_MILLIS = 20;
+  public static final long DEFAULT_POLL_TIMEOUT_MILLIS = 20L;
   public static final long PUBLISHER_REFERENCE_CLEANUP_TIMEOUT_MILLIS = 300L;
 
   public RangePublisherTest() {
--- a/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
+++ b/tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
@@ -26,6 +26,8 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static org.testng.Assert.assertTrue;
 import static org.testng.Assert.fail;
 
@@ -36,6 +38,7 @@ public class TestEnvironment {
   private static final long DEFAULT_TIMEOUT_MILLIS = 100;
 
   private static final String DEFAULT_NO_SIGNALS_TIMEOUT_MILLIS_ENV = "DEFAULT_NO_SIGNALS_TIMEOUT_MILLIS";
+  private static final String DEFAULT_POLL_TIMEOUT_MILLIS_ENV = "DEFAULT_POLL_TIMEOUT_MILLIS_ENV";
 
   private final long defaultTimeoutMillis;
   private final long defaultPollTimeoutMillis;
@@ -50,9 +53,9 @@ public class TestEnvironment {
    * the implementation, but can in some cases result in longer time to
    * run the tests.
    * @param defaultTimeoutMillis default timeout to be used in all expect* methods
-   * @param defaultPollTimeoutMillis default amount of time to poll for events if {@code defaultTimeoutMillis} isn't
-   *                                 preempted by an asynchronous event.
    * @param defaultNoSignalsTimeoutMillis default timeout to be used when no further signals are expected anymore
+   * @param defaultPollTimeoutMillis default amount of time to poll for events if {@code defaultTimeoutMillis} isn't
+    *                                preempted by an asynchronous event.
    * @param printlnDebug         if true, signals such as OnNext / Request / OnComplete etc will be printed to standard output,
    */
   public TestEnvironment(long defaultTimeoutMillis, long defaultNoSignalsTimeoutMillis, long defaultPollTimeoutMillis,
@@ -129,7 +132,7 @@ public class TestEnvironment {
    *                     often helpful to pinpoint simple race conditions etc.
    */
   public TestEnvironment(boolean printlnDebug) {
-    this(envDefaultTimeoutMillis(), envDefaultNoSignalsTimeoutMillis(), printlnDebug);
+    this(envDefaultTimeoutMillis(), envDefaultNoSignalsTimeoutMillis(), envDefaultPollTimeoutMillis(), printlnDebug);
   }
 
   /**
@@ -159,6 +162,14 @@ public class TestEnvironment {
   }
 
   /**
+   * The default amount of time to poll for events if {@code defaultTimeoutMillis} isn't preempted by an asynchronous
+   * event.
+   */
+  public long defaultPollTimeoutMillis() {
+    return defaultPollTimeoutMillis;
+  }
+
+  /**
    * Tries to parse the env variable {@code DEFAULT_TIMEOUT_MILLIS} as long and returns the value if present OR its default value.
    *
    * @throws java.lang.IllegalArgumentException when unable to parse the env variable
@@ -185,6 +196,21 @@ public class TestEnvironment {
       return Long.parseLong(envMillis);
     } catch (NumberFormatException ex) {
       throw new IllegalArgumentException(String.format("Unable to parse %s env value [%s] as long!", DEFAULT_NO_SIGNALS_TIMEOUT_MILLIS_ENV, envMillis), ex);
+    }
+  }
+
+  /**
+   * Tries to parse the env variable {@code DEFAULT_POLL_TIMEOUT_MILLIS_ENV} as long and returns the value if present OR its default value.
+   *
+   * @throws java.lang.IllegalArgumentException when unable to parse the env variable
+   */
+  public static long envDefaultPollTimeoutMillis() {
+    final String envMillis = System.getenv(DEFAULT_POLL_TIMEOUT_MILLIS_ENV);
+    if (envMillis == null) return envDefaultTimeoutMillis();
+    else try {
+      return Long.parseLong(envMillis);
+    } catch (NumberFormatException ex) {
+      throw new IllegalArgumentException(String.format("Unable to parse %s env value [%s] as long!", DEFAULT_POLL_TIMEOUT_MILLIS_ENV, envMillis), ex);
     }
   }
 
@@ -309,7 +335,7 @@ public class TestEnvironment {
   }
 
   /**
-   * Waits for {@link TestEnvironment#defaultTimeoutMillis()} and then verifies that no asynchronous errors
+   * Waits for {@link TestEnvironment#defaultNoSignalsTimeoutMillis()} and then verifies that no asynchronous errors
    * were signalled pior to, or during that time (by calling {@code flop()}).
    */
   public void verifyNoAsyncErrors() {
@@ -551,10 +577,10 @@ public class TestEnvironment {
     }
 
     public <E extends Throwable> void expectErrorWithMessage(Class<E> expected, String requiredMessagePart) throws Exception {
-      expectErrorWithMessage(expected, Collections.singletonList(requiredMessagePart), env.defaultTimeoutMillis(), env.defaultPollTimeoutMillis);
+      expectErrorWithMessage(expected, Collections.singletonList(requiredMessagePart), env.defaultTimeoutMillis(), env.defaultPollTimeoutMillis());
     }
     public <E extends Throwable> void expectErrorWithMessage(Class<E> expected, List<String> requiredMessagePartAlternatives) throws Exception {
-      expectErrorWithMessage(expected, requiredMessagePartAlternatives, env.defaultTimeoutMillis(), env.defaultPollTimeoutMillis);
+      expectErrorWithMessage(expected, requiredMessagePartAlternatives, env.defaultTimeoutMillis(), env.defaultPollTimeoutMillis());
     }
 
     @SuppressWarnings("ThrowableResultOfMethodCallIgnored")
@@ -584,7 +610,7 @@ public class TestEnvironment {
     }
 
     public <E extends Throwable> E expectError(Class<E> expected, long timeoutMillis) throws Exception {
-      return expectError(expected, timeoutMillis, env.defaultPollTimeoutMillis);
+      return expectError(expected, timeoutMillis, env.defaultPollTimeoutMillis());
     }
 
     public <E extends Throwable> E expectError(Class<E> expected, String errorMsg) throws Exception {
@@ -592,7 +618,7 @@ public class TestEnvironment {
     }
 
     public <E extends Throwable> E expectError(Class<E> expected, long timeoutMillis, String errorMsg) throws Exception {
-      return expectError(expected, timeoutMillis, env.defaultPollTimeoutMillis, errorMsg);
+      return expectError(expected, timeoutMillis, env.defaultPollTimeoutMillis(), errorMsg);
     }
 
     public <E extends Throwable> E expectError(Class<E> expected, long totalTimeoutMillis, long pollTimeoutMillis) throws Exception {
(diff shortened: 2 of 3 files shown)


Scottmitch:
I have updated the CopyrightWaiver.txt as part of https://github.com/reactive-streams/reactive-streams-jvm/pull/450

ktoso:
Thanks Scott! Seems I'll be able to review this, yay! I'll give it a look now 👍

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -1025,22 +1072,44 @@ public void expectCompletion(long timeoutMillis, String errorMsg) throws Interru
...
+      long totalTimeoutRemainingMillis = totalTimeoutMillis;
+      long timeStampA = System.nanoTime();
+      long timeStampB;
+
+      for (;;) {
+        Thread.sleep(Math.min(pollTimeoutMillis, totalTimeoutRemainingMillis));
+
+        if (env.asyncErrors.isEmpty()) {
+          timeStampB = System.nanoTime();
+          totalTimeoutRemainingMillis =- timeStampB - timeStampA;
```
viktorklang: @Scottmitch nanotime() is not in millis.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -1025,22 +1072,44 @@ public void expectCompletion(long timeoutMillis, String errorMsg) throws Interru
...
+    @SuppressWarnings("unchecked")
+    final <E extends Throwable> E expectError(Class<E> clazz, final long totalTimeoutMillis,
+                                              long pollTimeoutMillis,
+                                              String errorMsg) throws Exception {
+      long totalTimeoutRemainingMillis = totalTimeoutMillis;
+      long timeStampA = System.nanoTime();
+      long timeStampB;
+
+      for (;;) {
+        Thread.sleep(Math.min(pollTimeoutMillis, totalTimeoutRemainingMillis));
```
viktorklang: Using ns resolution we'll need to either use https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#sleep-long-int- or `LockSupport.parkNanos()`

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -39,6 +38,7 @@
   private static final String DEFAULT_NO_SIGNALS_TIMEOUT_MILLIS_ENV = "DEFAULT_NO_SIGNALS_TIMEOUT_MILLIS";
```
ktoso: Similar area to this PR, the verifyNoAsyncErrors has wrong docs actually, could you fix this along in the PR?

Should be

```java
  /**
   * Waits for {@link TestEnvironment#defaultNoSignalsTimeoutMillis()} and then verifies that no asynchronous errors
   * were signalled pior to, or during that time (by calling {@code flop()}).
   */
  public void verifyNoAsyncErrors() {
    verifyNoAsyncErrors(defaultNoSignalsTimeoutMillis());
  }
```

the comment now is:

`   * Waits for {@link TestEnvironment#defaultTimeoutMillis()} and then verifies that no asynchronous errors`

viktorklang:
Thanks for opening this PR @Scottmitch!

@reactive-streams/contributors Please chime in!

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -519,42 +551,57 @@ public void expectCompletion(long timeoutMillis, String errorMsg) throws Interru
     }
 
     public <E extends Throwable> void expectErrorWithMessage(Class<E> expected, String requiredMessagePart) throws Exception {
-      expectErrorWithMessage(expected, requiredMessagePart, env.defaultTimeoutMillis());
+      expectErrorWithMessage(expected, Collections.singletonList(requiredMessagePart), env.defaultTimeoutMillis(), env.defaultPollTimeoutMillis);
```
ktoso: If we end up introducing this new value, then it must not be accessed directly like this; 
it would have to gain an 

```
  public long defaultNoSignalsTimeoutMillis() {
    return defaultNoSignalsTimeoutMillis;
  }
```

Similar to  `defaultNoSignalsTimeoutMillis()` because that's a customization point people can use to override and set a value they may want to use there rather via the constructor;

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -1025,22 +1072,44 @@ public void expectCompletion(long timeoutMillis, String errorMsg) throws Interru
...
-        // ok, there was an expected error
-        Throwable thrown = env.asyncErrors.remove(0);
+      return expectError(clazz, timeoutMillis, timeoutMillis, errorMsg);
+    }
 
-        if (clazz.isInstance(thrown)) {
-          return (E) thrown;
+    @SuppressWarnings("unchecked")
+    final <E extends Throwable> E expectError(Class<E> clazz, final long totalTimeoutMillis,
+                                              long pollTimeoutMillis,
```
ktoso: nitpick: consistency on the `final` in the params; since its `long`s let's keep the signature similar-er as it was: no final on the `final long totalTimeoutMillis` as it cannot be modified anyway

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -1025,22 +1072,44 @@ public void expectCompletion(long timeoutMillis, String errorMsg) throws Interru
...
+                                              String errorMsg) throws Exception {
+      long totalTimeoutRemainingMillis = totalTimeoutMillis;
+      long timeStampA = System.nanoTime();
```
ktoso: As Viktor noted, there's an issue with the units here; so either Thread.sleep and `ms` (`System.currentTimeMillis()`) or sticking to nanos all the way. To be honest `ms` are likely good enough here, WDYT?

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -519,42 +551,57 @@ public void expectCompletion(long timeoutMillis, String errorMsg) throws Interru
     }
 
     public <E extends Throwable> void expectErrorWithMessage(Class<E> expected, String requiredMessagePart) throws Exception {
-      expectErrorWithMessage(expected, requiredMessagePart, env.defaultTimeoutMillis());
+      expectErrorWithMessage(expected, Collections.singletonList(requiredMessagePart), env.defaultTimeoutMillis(), env.defaultPollTimeoutMillis);
     }
     public <E extends Throwable> void expectErrorWithMessage(Class<E> expected, List<String> requiredMessagePartAlternatives) throws Exception {
-      expectErrorWithMessage(expected, requiredMessagePartAlternatives, env.defaultTimeoutMillis());
+      expectErrorWithMessage(expected, requiredMessagePartAlternatives, env.defaultTimeoutMillis(), env.defaultPollTimeoutMillis);
```
ktoso: `defaultPollTimeoutMillis()` and elsewhere too

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -50,15 +50,47 @@
...
+   * run the tests.
+   *
+   * @param defaultTimeoutMillis default timeout to be used in all expect* methods
+   * @param defaultNoSignalsTimeoutMillis default timeout to be used when no further signals are expected anymore
+   * @param defaultPollTimeoutMillis default amount of time to poll for events if {@code defaultTimeoutMillis} isn't
+   *                                 preempted by an asynchronous event.
+   */
+  public TestEnvironment(long defaultTimeoutMillis, long defaultNoSignalsTimeoutMillis, long defaultPollTimeoutMillis) {
+      this(defaultTimeoutMillis, defaultNoSignalsTimeoutMillis, defaultPollTimeoutMillis, false);
+  }
```
ktoso: the constructor dance is not yet pushed all the way; we have to implement being able to pick up the value from ENV, as people sometimes rely on doing this on CI etc.

So you'll need 

```java
  /**
   * Tries to parse the env variable {@code DEFAULT_POLL_TIMEOUT_MILLIS_ENV} as long and returns the value if present OR its default value.
   *
   * @throws java.lang.IllegalArgumentException when unable to parse the env variable
   */
  public static long envDefaultPollTimeoutMillis() {
    final String envMillis = System.getenv(DEFAULT_POLL_TIMEOUT_MILLIS_ENV);
    if (envMillis == null) return envDefaultTimeoutMillis();
    else try {
      return Long.parseLong(envMillis);
    } catch (NumberFormatException ex) {
      throw new IllegalArgumentException(String.format("Unable to parse %s env value [%s] as long!", DEFAULT_POLL_TIMEOUT_MILLIS_ENV, envMillis), ex);
    }
  }

```

and use it in:

```java
  public TestEnvironment(boolean printlnDebug) {
    this(envDefaultTimeoutMillis(), envDefaultPollTimeoutMillis(), envDefaultNoSignalsTimeoutMillis(), printlnDebug);
  }
```

And also

```java
  public TestEnvironment(long defaultTimeoutMillis, long defaultNoSignalsTimeoutMillis, boolean printlnDebug) {
    this(defaultTimeoutMillis, envDefaultPollTimeoutMillis(), defaultNoSignalsTimeoutMillis, printlnDebug);
  }
```

(yeah I know it's such a pain... wish Java would gain default params already :P)

Let's also reorder the constructor to be:

```java
  public TestEnvironment(long defaultTimeoutMillis, 
                         long defaultPollTimeoutMillis,
                         long defaultNoSignalsTimeoutMillis, 
                         boolean printlnDebug) {
```

sine the poll is more related to the timeout than the no signals I think (order of params must be the same _everywhere_, declarations, docs, fields).

ktoso:
The new config has to be able to be configured from the env variables, as some people rely on those -- commented in line how to do it.

And the param has to be documented as well in the [tck/README.md](https://github.com/reactive-streams/reactive-streams-jvm/blob/master/tck/README.md)

```
Note that the TCK differentiates between timeouts for "waiting for a signal" (`defaultTimeoutMillis`),
and "asserting no signals happen during a given amount of time" (`defaultNoSignalsTimeoutMillis`).
While the latter defaults to the prior, it may be useful to tweak them independently when running on continuous integration servers (for example, keeping the no-signals timeout significantly lower).

In order to configure these timeouts (for example when running on a slow continuous integration machine), you can either:

**Use env variables** to set these timeouts, in which case the you can do:

    export DEFAULT_TIMEOUT_MILLIS=100
    export DEFAULT_NO_SIGNALS_TIMEOUT_MILLIS=100
    export DEFAULT_POLL_TIMEOUT_MILLIS=20
    export PUBLISHER_REFERENCE_GC_TIMEOUT_MILLIS=300
```

Along the way there's some typos there we could fix (the above text should be fine to copy verbatim if you like it @Scottmitch :)

Sadly we have the readme in two places, since JDK and RS versions, so this would also have to be done in: https://github.com/reactive-streams/reactive-streams-jvm/tree/master/tck-flow/README.md

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -81,7 +113,7 @@ public TestEnvironment(long defaultTimeoutMillis, long defaultNoSignalsTimeoutMi
    * @param defaultTimeoutMillis default timeout to be used in all expect* methods
    */
   public TestEnvironment(long defaultTimeoutMillis) {
-    this(defaultTimeoutMillis, defaultTimeoutMillis, false);
+    this(defaultTimeoutMillis, defaultTimeoutMillis, defaultTimeoutMillis);
```
ktoso: not sure if we need to default the new option to the same as it was before -- perhaps we can default it to a fourth of it or something? Note also the logic about getting the value from env https://github.com/reactive-streams/reactive-streams-jvm/pull/452/files#r278075241

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -50,15 +50,47 @@
...
+   * @param defaultPollTimeoutMillis default amount of time to poll for events if {@code defaultTimeoutMillis} isn't
+   *                                 preempted by an asynchronous event.
    * @param defaultNoSignalsTimeoutMillis default timeout to be used when no further signals are expected anymore
    * @param printlnDebug         if true, signals such as OnNext / Request / OnComplete etc will be printed to standard output,
    */
-  public TestEnvironment(long defaultTimeoutMillis, long defaultNoSignalsTimeoutMillis, boolean printlnDebug) {
+  public TestEnvironment(long defaultTimeoutMillis, long defaultNoSignalsTimeoutMillis, long defaultPollTimeoutMillis,
+                         boolean printlnDebug) {
     this.defaultTimeoutMillis = defaultTimeoutMillis;
+    this.defaultPollTimeoutMillis = defaultPollTimeoutMillis;
```
ktoso: I think we should require (`if ... else throw`) that `defaultPollTimeoutMillis` be equal or smaller than `defaultTimeoutMillis`, otherwise we could get into some weird situations.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -1025,22 +1072,44 @@ public void expectCompletion(long timeoutMillis, String errorMsg) throws Interru
       } // else, ok
     }
 
-    @SuppressWarnings("unchecked")
+    /**
+     * @deprecated Deprecated in favor of {@link #expectError(Class, long, long, String)}.
+     */
+    @Deprecated
     public <E extends Throwable> E expectError(Class<E> clazz, long timeoutMillis, String errorMsg) throws Exception {
```
ktoso: deprecating sounds good; noone should have really touched those methods outside of the TCK itself I suppose... though yeah, let's play nice and not remove the method :)

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -519,42 +551,57 @@ public void expectCompletion(long timeoutMillis, String errorMsg) throws Interru
...
+                      err.getClass(), "anyOf: " + requiredMessagePartAlternatives, err.getMessage()));
     }
 
     public <E extends Throwable> E expectError(Class<E> expected) throws Exception {
       return expectError(expected, env.defaultTimeoutMillis());
     }
 
     public <E extends Throwable> E expectError(Class<E> expected, long timeoutMillis) throws Exception {
-      return expectError(expected, timeoutMillis, String.format("Expected onError(%s)", expected.getName()));
+      return expectError(expected, timeoutMillis, env.defaultPollTimeoutMillis);
```
ktoso: `env.defaultPollTimeoutMillis()`

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -39,6 +38,7 @@
   private static final String DEFAULT_NO_SIGNALS_TIMEOUT_MILLIS_ENV = "DEFAULT_NO_SIGNALS_TIMEOUT_MILLIS";
 
   private final long defaultTimeoutMillis;
+  private final long defaultPollTimeoutMillis;
```
ktoso: I wonder if the name shouldn't be `defaultErrorPollTimeoutMillis` as it does _not_ really touch any of the next/completion queue things, as those always used the ABQ: `      Optional<T> value = abq.poll(timeoutMillis, TimeUnit.MILLISECONDS);` so there's no change to that behavior by the introduction of this new param... WDYT @Scottmitch ?

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -50,15 +50,47 @@
    * the implementation, but can in some cases result in longer time to
    * run the tests.
    * @param defaultTimeoutMillis default timeout to be used in all expect* methods
+   * @param defaultPollTimeoutMillis default amount of time to poll for events if {@code defaultTimeoutMillis} isn't
```
ktoso: Note that it only applies for errors as other signals are caught whenever they appear?
Refs https://github.com/reactive-streams/reactive-streams-jvm/pull/452/files#r278079420

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: good catch

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: The intention was not to sleep with NS resolution. IMHO millis should be sufficient for this use case as it is an approximation anyways (and used for testing).

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: done

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: I can add a new accessor for `defaultPollTimeoutMillis` if you prefer, note that the accessor was not added to avoid increasing the footprint of this class.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: The `final`ness isn't meant to highlight that the caller's value won't be modified (primitives are passed by value as you eluded to), but to make sure this method preserves the value because it is used in the exception when a timeout occurs and the original value is expected. The `final` is meant to highlight the original value should be preserved as opposed to using it for intermediate calculations (e.g. instead of the `totalTimeoutRemainingMillis`  temp copy). I can remove it if you prefer because functionally it would be equivalent with the current code.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: yes, the unit mismatch will be fixed. ms sleep duration is sufficient, however time elapsed should use `nanoTime()` (due to monotonic properties of the API).

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: done

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: > And also...

The intention was to preserve backwards compatibility/behavior in this PR. If we grab the value from the environment here we will no longer be preserving existing behavior. Are you sure you want

> Let's also reorder the constructor to be:

The relative ordering of arguments would be different than other constructors.

This already exists...
```java
public TestEnvironment(long defaultTimeoutMillis, long defaultNoSignalsTimeoutMillis)
```

so we cannot have:
```java
public TestEnvironment(long defaultTimeoutMillis, long defaultPollTimeoutMillis)
```

This would make for a subtle issue in transitioning to the following constructor (the types are compatible but the argument order/semantics swap):
```java
public TestEnvironment(long defaultTimeoutMillis, long defaultNoSignalsTimeoutMillis, long defaultPollTimeoutMillis)
```

I can re-order if you prefer but not sure if this is a clear win (considering the existing api/behavior).

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: I can do this if you prefer, however we need to take the `min` between "remaining total time" and "poll time" anyways so this restriction is not necessary. For consistency we would also have to enforce it in `expectError` which may pop-up in unexpected cases.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: yip ... bunch of methods look over exposed here, but I don't want to break anything.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: I'll defer to you here, but I think it would be surprising/undesirable to change existing behavior of a 1.x project.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: The rational was the `error` handling is an implementation detail and this polling can be used for all things that would otherwise have to block/wait the full `defaultTimeoutMillis` duration (e.g. if there is no async pre-emption). Although it is currently used for expected error timeouts due to internal implementation details, if the implementation was to change or be extended I think it is useful to have a "poll interval" and "upper bound" for these cases. However I'll defer to you if you want it more specialized.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: The rational was the `error` handling is an implementation detail and this polling can be used for all things that would otherwise have to block/wait the full `defaultTimeoutMillis` duration (e.g. if there is no async pre-emption). Although it is currently used for expected error timeouts due to internal implementation details, if the implementation was to change or be extended I think it is useful to have a "poll interval" and "upper bound" for these cases. However I'll defer to you if you want it more specialized.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: done

tck-flow/README.md
```diff
@@ -209,32 +209,39 @@ within the TCK which await for something to happen. The other timeout is `publis
...
 Or **define the timeouts explicitly in code**:
 
 ```java
-public class RangePublisherTest extends FlowPublisherVerification<Integer> {
+public class RangePublisherTest extends PublisherVerification<Integer> {
 
   public static final long DEFAULT_TIMEOUT_MILLIS = 100L;
   public static final long DEFAULT_NO_SIGNALS_TIMEOUT_MILLIS = DEFAULT_TIMEOUT_MILLIS;
-  public static final long PUBLISHER_REFERENCE_CLEANUP_TIMEOUT_MILLIS = 500L;
+  public static final long DEFAULT_POLL_TIMEOUT_MILLIS = 20;
```
ktoso: sorry, mini nitpick: add L for consistency

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
ktoso: Ok that's a good point, let's keep the general wording for "whenever would have to poll" 👍

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -309,7 +335,7 @@ public Throwable dropAsyncError() {
   }
 
   /**
-   * Waits for {@link TestEnvironment#defaultTimeoutMillis()} and then verifies that no asynchronous errors
+   * Waits for {@link TestEnvironment#defaultNoSignalsTimeoutMillis()} and then verifies that no asynchronous errors
```
ktoso: Thanks!

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
ktoso: My thinking was that the only place this would change is that it may poll a few times and catch an error earlier... shouldn't really impact in other ways.

True through, we can stick to what it was and people can customize if they want to 🤔 Let's do that I guess, as you said.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
ktoso: Oh true, the `min` handles it there quite nicely. Seems like a good idea to rely on that and not push more throws in there.

It could cause an issue if someone used the `expectError` with passed in timeout values explicitly -- noone really should have, but can't know if they did 🤔 

Sounds good to keep as is 👍

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
ktoso: Hah, another good point about the compat here... Ignore me here, was looking at it from a too "api puristic" point of view there.

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
```diff
@@ -156,6 +199,21 @@ public static long envDefaultNoSignalsTimeoutMillis() {
     }
   }
 
+  /**
+   * Tries to parse the env variable {@code DEFAULT_POLL_TIMEOUT_MILLIS_ENV} as long and returns the value if present OR its default value.
+   *
+   * @throws java.lang.IllegalArgumentException when unable to parse the env variable
+   */
+  public static long envDefaultPollTimeoutMillis() {
```
ktoso: great thanks 👍

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
ktoso: I see, thanks for the explanation, let's keep then 👍

tck/src/main/java/org/reactivestreams/tck/TestEnvironment.java
Scottmitch: ok I will keep as is. fwiw I agree your suggestion is an improvement from an API perspective 👍

Scottmitch:
thanks for review @ktoso and @viktorklang! Any other comments, or can this be merged?

tck-flow/README.md
Scottmitch: done

ktoso:
Looks good as I said above here. Ping @viktorklang. Sadly I never got merge powers here... ¯\_(ツ)_/¯

viktorklang:
@Scottmitch Thanks for the great contribution!

@ktoso Thanks for helping out reviewing, Konrad! ^^
I'll investigate if we can give you merge-powers, it seems like an oversight.

Issue #456, opened 2019-05-27
Make it possible to publish the RS jars locally


viktorklang:
@reactive-streams/contributors & community: if anybody here has experience with Gradle and could fix this issue, that would be awesome. Simply adding the maven-publish plugin and do publishToMavenLocal doesn't seem to work.

PR #461, opened 2019-07-03
Enable ./gradlew publishToMavenLocal

Scottmitch:
Motivation:
Add a publishing task in the top level gradle file that allows for publishing
to the local maven repository.

Modifications:
- When `./gradlew publishToMavenLocal` is invoked we should publish to the local
maven repo for: reactive-streams, reactive-stream-tck,
reactive-streams-tck-flow, and reactive-streams-examples

Result:
`./gradlew publishToMavenLocal` publishes artifacts to the local maven
repository.
Fixes https://github.com/reactive-streams/reactive-streams-jvm/issues/456

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,6 +52,7 @@ subprojects {
                  "reactive-streams-examples"]) {
         apply plugin: "maven"
         apply plugin: "signing"
+        apply plugin: "maven-publish"
 
         signing {
             sign configurations.archives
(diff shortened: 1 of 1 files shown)


Scottmitch:
I have verified this works locally, but I'm no gradle expert and can't be sure this won't have other side effects. Reviews/comments/suggestions are welcome.

viktorklang:
@Scottmitch LGTM! Merging :)

PR #462, opened 2019-07-03
Allow triggerRequest and signalCancel to be coalesced

jroper:
See #453 and https://github.com/reactive-streams/reactive-streams-jvm/pull/441#discussion_r288380996

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
@@ -266,8 +266,9 @@ public abstract class SubscriberWhiteboxVerification<T> extends WithHelperPublis
       @Override
       public void run(WhiteboxTestStage stage) throws InterruptedException {
         stage.puppet().triggerRequest(1);
-        stage.puppet().signalCancel();
         stage.expectRequest();
+        stage.puppet().signalCancel();
+        stage.expectCancelling();
         stage.signalNext();
 
         stage.puppet().triggerRequest(1);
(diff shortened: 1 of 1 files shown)


tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
```diff
@@ -266,8 +266,9 @@ public void required_spec208_mustBePreparedToReceiveOnNextSignalsAfterHavingCall
       @Override
       public void run(WhiteboxTestStage stage) throws InterruptedException {
         stage.puppet().triggerRequest(1);
-        stage.puppet().signalCancel();
         stage.expectRequest();
+        stage.puppet().signalCancel();
+        stage.expectCancelling();
```
jroper: Now we expect the request before signalling cancel, preventing the coalescing of the request and cancel, and by expecting cancelling before signalling next, we ensure that signalling next does test what this test says it tests.

jrudolph: Sounds reasonable and less prone to race conditions.

viktorklang:
@reactive-streams/contributors Feedback on this?

akarnokd:
I'm not sure. Is this something because of Akka Stream's sequential signal count and mixing requests with cancellation in the same actor message queue?

tck/src/main/java/org/reactivestreams/tck/SubscriberWhiteboxVerification.java
Scottmitch: +1. the behavior you described sounds like an improvement.

viktorklang:
@akarnokd I think @jroper will be able to answer that. That aside though, the importance to the TCK is to be able to enforce the spec, so it is more about making sure of that.

viktorklang:
@jroper @akarnokd @reactive-streams/contributors Let's make some progress here so we can release the first 1.0.3 RC. Feedback encouraged

Scottmitch:
@viktorklang - can we merge this, cut an RC, and collect comments when folks kick the tires on the RC?

viktorklang:
@Scottmitch Agreed—people have had ample time to respond as is. Will tackle this on Monday

PR #473, opened 2019-12-03
Upgrade to Gradle 6.0.1, TestNG 7.0.0, use Bnd for OSGi properties

akarnokd:
Upgrading to Gradle 6 requires using the Bnd plugin and a slightly different way to define OSGi parameters. 

In addition, Gradle complains about `compile` being deprecated and thus have been upgraded to `api`/`implementation` use. 

Third, TestNG has been updated to the latest because that 5.10.4 won't run under Eclipse as its plugin stopped supporting 5 some time ago afaik. In consequence, one of the methods had to be annotated as non-testable because TestNG tries to inject parameters into public methods of a test-annotated class.


api/build.gradle
```diff
@@ -19,7 +19,14 @@ sourceSets {
...
 jar {
-    manifest {
-        attributes('Automatic-Module-Name': 'org.reactivestreams')
-    }
+    bnd ('Bundle-Name': 'reactive-streams-jvm',
+            'Bundle-Vendor': 'Reactive Streams SIG',
+            'Bundle-Description': 'Reactive Streams API',
+            'Bundle-DocURL': 'http://reactive-streams.org',
+            'Bundle-Version': project.version,
+            'Import-Package': '!org.junit,!junit.framework,!org.mockito.*,!org.testng.*,*',
```
viktorklang: Are these needed for all the jars?

akarnokd: It used to be an import bloat with OSGi due to it including test dependencies. I'll double check.

akarnokd: Yep, they are no longer necessary, plus the TestNG imports have to be named for the TCK bundle anyway.

viktorklang:
Thanks for fixing this, @akarnokd! Very appreciated :)

PR #475, opened 2020-01-17
Fix a typo in spec101 error message

akarnokd:
A missing space.


viktorklang:
Nice catch, @akarnokd—thanks for the assistance! :+1:

PR #476, opened 2020-01-19
example implementation of Publisher added - ThreadPublisher.

akaigoro:
This is the most compact implementation of the org.reactivestreams.Publisher interface, and as such can be useful for students learning reactive programming.


akarnokd:
Well, no.

> The purpose of Reactive Streams is to provide a standard for asynchronous stream processing with **non-blocking backpressure**.

PR #483, opened 2020-02-27
Fix TCK §2.3 verification for when a legit call is made with onError/onComplete method name in the stacktrace

### LINKED ISSUE #460: (TCK) SubscriberBlackboxVerification.required_spec203_blackbox_mustNotCallMethodsOnSubscriptionOrPublisherInOnComplete (onError)
There are two tests in `SubscriberBlackboxVerification` that verify conformance to Rule 2.3:

> `Subscriber.onComplete()` and `Subscriber.onError(Throwable t)` MUST NOT call any methods on the `Subscription` or the `Publisher`.

I can see that this is done by examining the call stack, looking for the specific methods' names:
```java
        final Subscription subs = new Subscription() {
          @Override
          public void request(long n) {
            Throwable thr = new Throwable();
            for (StackTraceElement stackElem : thr.getStackTrace()) {
              if (stackElem.getMethodName().equals("onError")) {
                env.flop(String.format("Subscription::request MUST NOT be called from Subscriber::onError (Rule 2.3)! (Caller: %s::%s line %d)",
                                       stackElem.getClassName(), stackElem.getMethodName(), stackElem.getLineNumber()));
              }
            }
          }

          @Override
          public void cancel() {
            Throwable thr = new Throwable();
            for (StackTraceElement stackElem : thr.getStackTrace()) {
              if (stackElem.getMethodName().equals("onError")) {
                env.flop(String.format("Subscription::cancel MUST NOT be called from Subscriber::onError (Rule 2.3)! (Caller: %s::%s line %d)",
                                       stackElem.getClassName(), stackElem.getMethodName(), stackElem.getLineNumber()));
              }
            }
          }
        };
```

It seems to do the job, however, the approach might be brittle. The reason is that approach is not specific enough. It might produce false positives as it disregards the objects on which `onError` and `onComplete` are called. A method can be similarly named, but otherwise unrelated. Or the method can be called on a different `Subscriber` (i.e. calling `subscription1.cancel()` from `subscriber2.onComplete()`).

Could a `ThreadLocal<Boolean>` flag solution be any better? Before calling `onError` on `onComplete` on `Subscriber` the test sets that flag. Upon exit from those methods the flag is cleared. Inside `Subscription`'s `cancel` and `request(long)` the flag is being checked.

akarnokd:
Originally, the TCK verifying rule §2.3 would fail if the call stack contained any method named `onComplete` or `onError`, even if they were not on the call chain initiated by the test's `onError`/`onComplete`.

Related #481, https://github.com/eclipse/microprofile-reactive-streams-operators/issues/131
Resolves #460

### DIFF
--- a/tck/src/main/java/org/reactivestreams/tck/SubscriberBlackboxVerification.java
+++ b/tck/src/main/java/org/reactivestreams/tck/SubscriberBlackboxVerification.java
@@ -128,7 +128,7 @@ public abstract class SubscriberBlackboxVerification<T> extends WithHelperPublis
     blackboxSubscriberWithoutSetupTest(new BlackboxTestStageTestRun() {
       @Override
       public void run(BlackboxTestStage stage) throws Throwable {
-        final String onCompleteMethod = "required_spec203_mustNotCallMethodsOnSubscriptionOrPublisherInOnComplete_call";
+        final String onCompleteMethod = "required_spec203_blackbox_mustNotCallMethodsOnSubscriptionOrPublisherInOnComplete_call";
 
         final Subscription subs = new Subscription() {
           @Override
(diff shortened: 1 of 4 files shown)


danielkec:
Should also fix #460

viktorklang:
Thanks @akarnokd! :+1:

@reactive-streams/contributors LGTM or not?

viktorklang:
Thanks @akarnokd!

Issue #484, opened 2020-04-03
Doc unification (styling, minor)


tomislavhofman:
Docs written in `Processor`, `Publisher`, `Subscriber` and `Subscription` interfaces do not conform to the same styling across the files.

PR #485, opened 2020-04-03
Styling unification.

tomislavhofman:
Related to #484

### DIFF
--- a/api/src/main/java/org/reactivestreams/Processor.java
+++ b/api/src/main/java/org/reactivestreams/Processor.java
@@ -15,8 +15,8 @@ package org.reactivestreams;
  * A {@link Processor} represents a processing stage—which is both a {@link Subscriber}
  * and a {@link Publisher} and obeys the contracts of both.
  *
- * @param <T> the type of element signaled to the {@link Subscriber}.
- * @param <R> the type of element signaled by the {@link Publisher}.
+ * @param <T> the type of element signaled to the {@link Subscriber}
+ * @param <R> the type of element signaled by the {@link Publisher}
  */
 public interface Processor<T, R> extends Subscriber<T>, Publisher<R> {
 }
--- a/api/src/main/java/org/reactivestreams/Processor.java
+++ b/api/src/main/java/org/reactivestreams/Processor.java
@@ -12,11 +12,11 @@
 package org.reactivestreams;
 
 /**
- * A Processor represents a processing stage—which is both a {@link Subscriber}
+ * A {@link Processor} represents a processing stage—which is both a {@link Subscriber}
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/reactivestreams/Publisher.java
```diff
@@ -15,7 +15,7 @@
  * A {@link Publisher} is a provider of a potentially unbounded number of sequenced elements, publishing them according to
  * the demand received from its {@link Subscriber}(s).
  * <p>
- * A {@link Publisher} can serve multiple {@link Subscriber}s subscribed {@link #subscribe(Subscriber)} dynamically
+ * A {@link Publisher} can serve multiple {@link Subscriber}(s) subscribed {@link Publisher#subscribe(Subscriber)} dynamically
```
akarnokd, line 18: `multiple` already indicates plural here so `s` is fine.

api/src/main/java/org/reactivestreams/Processor.java
```diff
@@ -12,11 +12,11 @@
...
 
 /**
- * A Processor represents a processing stage—which is both a {@link Subscriber}
+ * A {@link Processor} represents a processing stage—which is both a {@link Subscriber}
  * and a {@link Publisher} and obeys the contracts of both.
  *
- * @param <T> the type of element signaled to the {@link Subscriber}
- * @param <R> the type of element signaled by the {@link Publisher}
+ * @param <T> the type of element signaled to the {@link Subscriber}.
+ * @param <R> the type of element signaled by the {@link Publisher}.
```
akarnokd, line 19: Such parameter descriptions are not full sentences themselves so I don't think `.` is needed here and elsewhere.

api/src/main/java/org/reactivestreams/Publisher.java
tomislavhofman: Oops missed that, good catch.

api/src/main/java/org/reactivestreams/Processor.java
tomislavhofman: Alright, I'll remove those!

viktorklang:
Hi @tomislavhofman,

Thanks for helping out with the documentation, much appreciated!

Would you also make sure you're added to the COPYRIGHTWAIVERS file?

√

tomislavhofman:
Done it under first commit,
apologies for not making it clear!

viktorklang:
@tomislavhofman No problem! (I didn't see it in the file list first for some reason)

viktorklang:
Thanks @tomislavhofman!

tomislavhofman:
Glad to contribute!

PR #488, opened 2020-05-07
Clarify intent of rule 1.9

### LINKED ISSUE #486: Happens-before relationship between `Publisher#subscribe()` and `Subscriber.onSubscribe()`?
Considering a simple example:

```java
public class UnsafeSubscriber implements Subscriber<String> {
    private boolean duplicateOnSubscribe = false;

    @Override
    public void onSubscribe(final Subscription s) {
        if (duplicateOnSubscribe) {
            throw new IllegalStateException("Duplicate onSubscribe() calls.");
        }
        duplicateOnSubscribe = true;
    }

    @Override
    public void onNext(final String s) {

    }

    @Override
    public void onError(final Throwable t) {

    }

    @Override
    public void onComplete() {

    }
}
```
If an `UnsafeSubscriber` instance is created in a different thread than the one that invokes `onSubscribe()` (true for an asynchronous `Publisher`), according to the [java memory model](https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.5), this statement inside `onSubscribe()`:

`if (duplicateOnSubscribe) {` 

is guaranteed to compute to `false` if and only if the instance is published safely between these threads. None of the rules in the specifications establish a happens-before relationship between `Publisher#subscribe()` and `Subscriber#onSubscribe()`. So, the usage above can be categorized as unsafe. In a more convoluted form, the assignment:

`private boolean duplicateOnSubscribe = false;`

can be interleaved with 

`duplicateOnSubscribe = true;` such that `duplicateOnSubscribe` is set to `false` later.
 
Has this been considered before or am I missing something?

NiteshKant:
Rule 1.9 defines that `onSubscribe()` on a `Subscriber` should be called before calling any other method on that `Subscriber`. However, this does not clarify that if `onSubscribe()` is signalled asynchronously then there should be a happens-before relationship between `subscribe()` and signalling of `onSubscribe()`.
In absence of such a guarantee, non-final fields initialized in a constructor have no guarantee to be visible inside `onSubscribe()`.

Considering a simple example:

public class UnsafeSubscriber implements Subscriber<String> {
    private boolean duplicateOnSubscribe = false;

    @Override
    public void onSubscribe(final Subscription s) {
        if (duplicateOnSubscribe) {
            throw new IllegalStateException("Duplicate onSubscribe() calls.");
        }
        duplicateOnSubscribe = true;
    }

    @Override
    public void onNext(final String s) {

    }

    @Override
    public void onError(final Throwable t) {

    }

    @Override
    public void onComplete() {

    }
}
If an UnsafeSubscriber instance is created in a different thread than the one that invokes onSubscribe() (true for an asynchronous Publisher), according to the java memory model, this statement inside onSubscribe():

if (duplicateOnSubscribe) {

is guaranteed to compute to false if and only if the instance is published safely between these threads. None of the rules in the specifications establish a happens-before relationship between Publisher#subscribe() and Subscriber#onSubscribe(). So, the usage above can be categorized as unsafe. In a more convoluted form, the assignment:

private boolean duplicateOnSubscribe = false;

can be interleaved with

duplicateOnSubscribe = true; such that duplicateOnSubscribe is set to false later.

Has this been considered before or am I missing something?

Fixes #486


viktorklang:
@NiteshKant Your Subscriber is not a valid subscriber since it violates 2:13: https://github.com/reactive-streams/reactive-streams-jvm#2.13

NiteshKant:
@viktorklang yikes, yes. Throwing `IllegalArgumentException` isn't the crux of the problem though. Once we get to an agreement on #486 about whether this PR make sense, I wil fix the commit message as:

```java
if (duplicateOnSubscribe) {
        System.out.println("Duplicate onSubscribe() calls.");
}
```

PR #489, opened 2020-05-19
clarify rule 2.7 intent

OlegDokuka:
The intent of this PR is to clearly specify that there is no possibility to have racing between two request calls. 

Having just happens-before relation between calls in the clarification of the rule may lead to the assumption that calls may overlap (which contradicts to serial behavior). I have observed that such misunderstanding may appear in one of the issues in rsocket-java, thus want to eliminate that unclearness in the spec (see https://github.com/rsocket/rsocket-java/issues/837).

Let me know if I'm treating that rule improperly.

Cheers,
Oleh


viktorklang:
Hi @OlegDokuka!

While the rule is clear about it having to be serial, I agree with you that it makes sense to repeat it in the explanation, since it has stricter implications than a "simple" happens-before.

viktorklang:
@reactive-streams/contributors Wdyt?

OlegDokuka:
Guessing nobody is against so we can merge that one. WDYT @viktorklang ?

viktorklang:
@OlegDokuka Merged, thank you!

OlegDokuka:
My pleasure!

PR #494, opened 2020-07-24
provides clarification on the 1.9 rule

### LINKED ISSUE #486: Happens-before relationship between `Publisher#subscribe()` and `Subscriber.onSubscribe()`?

OlegDokuka:
closes #486


OlegDokuka:
It seems that `SubmissionPublisher` violates rule 1.9 :D

DougLea:
@OlegDokuka Can you describe the failed test?

OlegDokuka:
The idea is to mandate calling `onSubscribe` within the `subscribe` method synchronously (on the caller thread). 

Due to my findings, `SubmissionPublisher` always offload that call to the executor, which makes that process asynchronous

akarnokd:
This is a drastic change and I don't understand why you'd want to add this limitation.

OlegDokuka:
@akarnokd I guess that was a drastic misunderstanding of how it should work. 

Imagin `.subscribe(subscriber)` offloads task to a scheduler which never executes a task or hangs for any reason. How to you cancel that one? 

The correct answer - nohow

akarnokd:
If this is a failure mode for your use case, don't use an operator that schedules onSubscribe. Also thete is a limit on how much infrastructure failures we can work aroud, i.e. can't do much when trying to schedule with a shut down executor for example.

I don't support this change to the spec/tck.

Instead let's look at the original issue and have a workaround/solution there.

OlegDokuka:
Well. Sounds like solving symptoms rather than solving the root cause. 
We have already fixed that. My intent is to get rid of such a mistake in the future.

And yeah - the workaround is simple - do not offload anything related to subscription phase

akarnokd:
Your solution breaks at least one established implementation that passed before. I'd also think this could break Akka Stream as they tend to schedule all sorts of signals.

OlegDokuka:
I'm not pushing to merge it right now. That discussion started at #486 and ended saying that we do not have to offload the `onSubscribe` call.

If we know that there is an extensive number of libraries using that - let's discuss how to bring that breaking clarification up

DougLea:
@OlegDokuka SubmissionPublisher calls Subscriber.onSubscribe synchronously. It does not trap exceptions from the call, so they will be rethrown. I suppose this could be changed, but whatever is done would be outside of spec because onSubscribe should not throw. See code at http://gee.cs.oswego.edu/cgi-bin/viewcvs.cgi/jsr166/src/main/java/util/concurrent/SubmissionPublisher.java?view=log

OlegDokuka:
@DougLea 

> SubmissionPublisher calls Subscriber.onSubscribe synchronously.

That is not true. `Subscriber#onSubscribe` is scheduled on the executor service. So the call is offloaded and asynchronous.

The execution stack is the following 

1. *Sync_ Thread 1*: [`SubmissionPublisher#subscribe(subscriber)`](http://gee.cs.oswego.edu/cgi-bin/viewcvs.cgi/jsr166/src/main/java/util/concurrent/SubmissionPublisher.java?view=markup#l313)
2. *Sync_ Thread 1*: [`BufferedSubscription#onSubscribe()`](http://gee.cs.oswego.edu/cgi-bin/viewcvs.cgi/jsr166/src/main/java/util/concurrent/SubmissionPublisher.java?view=markup#l331)
3. *Sync_ Thread 1*: [`BufferedSubscription#startOnSignal(RUN \| ACTIVE)`](http://gee.cs.oswego.edu/cgi-bin/viewcvs.cgi/jsr166/src/main/java/util/concurrent/SubmissionPublisher.java?view=markup#l1217)
4. *Sync_ Thread 1*: [`BufferedSubscription#tryStart()`](http://gee.cs.oswego.edu/cgi-bin/viewcvs.cgi/jsr166/src/main/java/util/concurrent/SubmissionPublisher.java?view=markup#l1213)
5. *Sync_ Thread 1*: [`Executor#execute`](http://gee.cs.oswego.edu/cgi-bin/viewcvs.cgi/jsr166/src/main/java/util/concurrent/SubmissionPublisher.java?view=markup#l1197). 
6. *Async Thread 2*: [`BufferedSubscription#subscribeOnOpen`](http://gee.cs.oswego.edu/cgi-bin/viewcvs.cgi/jsr166/src/main/java/util/concurrent/SubmissionPublisher.java?view=markup#l1263)
7. *Async Thread 2*: [`BufferedSubscription#consumeSubscribe`](http://gee.cs.oswego.edu/cgi-bin/viewcvs.cgi/jsr166/src/main/java/util/concurrent/SubmissionPublisher.java?view=markup#l1349)
8. *Async Thread 2*: [`Subscriber#onSubscribe(Subscription)`](http://gee.cs.oswego.edu/cgi-bin/viewcvs.cgi/jsr166/src/main/java/util/concurrent/SubmissionPublisher.java?view=markup#l1357)

**_Please, notice, at point 5, the tasks is offloaded to the `Executor` so the execution is not synchronous anymore_**.


>  It does not trap exceptions from the call, so they will be rethrown. I suppose this could be changed, but whatever is done would be outside of spec because onSubscribe should not throw

Nobody is talking about exceptions in `onSubscribe` method.

The PR here makes sure every `Publisher` implementation calls `onSubscribe` immediately (a.k.a synchronously) (a.k.a on the caller `Thread` within the same stack) within the `Publisher#subscriber` method.

It means that the following is ✅

```
Publisher {

 subscriber(Subscriber s) {
    s.onSubscribe(new Subscription()); ✅

    ...
 }

}
```

whiles the next is ❌


```
Publisher {

 subscriber(Subscriber s) {
    ForkJoinPool.execute(() -> s.onSubscribe(new Subscription())); ❌❌❌❌❌❌❌

    ...
 }

}
```

DougLea:
@OlegDokuka Here is the code. It looks synchro9nous to me:
    public void subscribe(Subscriber<? super T> subscriber) {
        if (subscriber == null) throw new NullPointerException();
        ReentrantLock lock = this.lock;
        int max = maxBufferCapacity; // allocate initial array
        Object[] array = new Object[max < INITIAL_CAPACITY ?
                                    max : INITIAL_CAPACITY];
        BufferedSubscription<T> subscription =
            new BufferedSubscription<T>(subscriber, executor, onNextHandler,
                                        array, max);
        lock.lock();
        try {
            if (!subscribed) {
                subscribed = true;
                owner = Thread.currentThread();
            }
            for (BufferedSubscription<T> b = clients, pred = null;;) {
                if (b == null) {
                    Throwable ex;
                    **subscription.onSubscribe();**
                    if ((ex = closedException) != null)
                        subscription.onError(ex);
                    else if (closed)
                        subscription.onComplete();
                    else if (pred == null)
                        clients = subscription;
                    else
                        pred.next = subscription;
                    break;
                }
                BufferedSubscription<T> next = b.next;
                if (b.isClosed()) {   // remove
                    b.next = null;    // detach
                    if (pred == null)
                        clients = next;
                    else
                        pred.next = next;
                }
                else if (subscriber.equals(b.subscriber)) {
                    b.onError(new IllegalStateException("Duplicate subscribe"));
                    break;
                }
                else
                    pred = b;
                b = next;
            }
        } finally {
            lock.unlock();
        }
    }

OlegDokuka:
@DougLea You have all the references to the execution stack, we can debate forever, but it is better to check it first before

DougLea:
@OlegDokuka Sorry for the confusion. You are right that user.onSubscribe method can be performed async.

Issue #528, opened 2021-03-05
Upgrade testng in reactive-streams-tck


barambani:
Currently reactive-streams-tck is bringing testng at version 7.0.0 that was using a jcommander version (`com.beust:jcommander@1.72 `) affected by a security vulnerability 

https://snyk.io/vuln/SNYK-JAVA-COMBEUST-174815

please consider updating it.

lukaseder:
I'm running into this as well:

```
13:29:51,262 [ERROR] One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '0.0':
13:29:51,268 [ERROR] bsh-2.0b4.jar: CVE-2016-2510
13:29:51,292 [ERROR] snakeyaml-1.6.jar: CVE-2017-18640
```

Is there a publication of a new version on the roadmap? 1.0.3 still has the dependency:
https://search.maven.org/artifact/org.reactivestreams/reactive-streams-tck

PR #529, opened 2021-03-06
Addresses #528 by upgrading testng to 7.3.0

viktorklang:
Seemed like the lowest resistance path was to upgrade TestNG itself.


viktorklang:
Upgrading to 7.4.0 seemed to create artifact resolution problems, which is weird since it definitely is on Maven Central, alas upgrading to 7.3.0 should be enough to make sure that jcommander is upgraded appropriately to 1.78: https://github.com/cbeust/testng/blob/7.3.0/build.gradle.kts#L59

akarnokd (approved):

PR #516, opened 2020-08-15
migrate from Travis CI to GitHub Actions CI

sullis:

PR #532, opened 2021-06-15
Do not build on JDK 13.

akarnokd:
Java 6 is no longer available as target in 13. Removing that build target.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [ '8', '11', '13' ]
+        java: [ '8', '11' ]
     steps:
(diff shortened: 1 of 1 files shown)


viktorklang:
Is this PR related to #516? If so, @rkuhn are you able to merge this fix?

rkuhn:
Yes, this should fix it—we may want to discuss for how long we want to keep JDK 6 compatibility.

rkuhn:
question regarding DCO: is this required?

viktorklang:
@rkuhn Yes, DCO is required by the Reactive Foundation (via the Linux
Foundation)

rkuhn:
In this case: @akarnokd please add a `--signoff` to your commit (as described behind the Details link for DCO)—and thanks for the quick response!

akarnokd:
@rkuhn done.

PR #544, opened 2023-03-07
TestEnvironment: Fix timeout arithmetic

Scottmitch:
Motivation:
The timeout calculation decrementing `totalTimeoutRemainingNs` is incorrect which results in premature timeouts and test failures.

### DIFF
--- a/examples/src/test/java/org/reactivestreams/example/unicast/AsyncRangePublisherTest.java
+++ b/examples/src/test/java/org/reactivestreams/example/unicast/AsyncRangePublisherTest.java
@@ -1,0 +1,215 @@
+/***************************************************
+ * Licensed under MIT No Attribution (SPDX: MIT-0) *
+ ***************************************************/
+package org.reactivestreams.example.unicast;
+
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+import org.reactivestreams.tck.PublisherVerification;
+import org.reactivestreams.tck.TestEnvironment;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
+
+@Test // Must be here for TestNG to find and run this, do not remove
+public class AsyncRangePublisherTest extends PublisherVerification<Integer> {
+    private static final int TERMINAL_DELAY_MS = 20;
+    private static final int DEFAULT_TIMEOUT_MS = 5000;
+    private static final int DEFAULT_POLL_INTERVAL_MS = TERMINAL_DELAY_MS / 2;
+    private ExecutorService e;
+    @BeforeClass
+    void before() { e = Executors.newCachedThreadPool(); }
+    @AfterClass
+    void after() { if (e != null) e.shutdown(); }
+
+    public AsyncRangePublisherTest() {
+        super(new TestEnvironment(DEFAULT_TIMEOUT_MS, 50, DEFAULT_POLL_INTERVAL_MS));
+    }
+
+    @Override
+    public Publisher<Integer> createPublisher(long elements) {
+        return new AsyncPublisher<Integer>(new RangePublisher(1, (int)elements), e);
+    }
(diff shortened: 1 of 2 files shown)


Scottmitch:
@viktorklang - are you still merging commits? are there other contributors (do you need more)?

Scottmitch:
@rkuhn @akarnokd - wdyt?

akarnokd:
Indeed the original code looks wrong but I don't remember tripping this failure case ever. Also isn't there a TCK test that verifies this timeout case?

rkuhn:
The fact that this obvious bug only surfaces now means that this TCK facility is not used correctly anywhere in the code — `long` underflow seems unlikely to occur here. I can’t offer further comments as the TCK was mainly conceived and implemented by @ktoso (and I haven’t used JVM languages for the past few years).

Scottmitch:
If errors are delivered synchronously (like in the TCK tests) and the unit under test is behaving as expected, it is unlikely this code will be hit (e.g. `env.asyncErrors.isEmpty()` will be `false`). We occasionally hit the condition in ServiceTalk due to testing async signal delivery (different thread than test thread).

Scottmitch:
I added a unit test to the TCK to demonstrate the issue (if you remove the fix the test will likely fail).

examples/src/test/java/org/reactivestreams/example/unicast/AsyncRangePublisherTest.java
```diff
@@ -0,0 +1,175 @@
...
+        public void subscribe(Subscriber<? super T> s) {
+            original.subscribe(new AsyncSubscriber<T>(requireNonNull(s), executor));
+        }
+
+        private static final class AsyncSubscriber<T> implements Subscriber<T> {
+            private final BlockingQueue<Object> signalQueue = new LinkedBlockingQueue<Object>();
+
+            private AsyncSubscriber(final Subscriber<? super T> original, final Executor executor) {
+                try {
+                    executor.execute(new Runnable() {
```
akarnokd, line 62: Please don't let `this` escape the constructor like this.

akarnokd (changes requested):

examples/src/test/java/org/reactivestreams/example/unicast/AsyncRangePublisherTest.java
Scottmitch: good catch, fixed.

akarnokd (approved):

Scottmitch:
@akarnokd - thanks for approval. What are the next steps to getting this merged and released?

akarnokd:
I don't know. Most of the original crew have moved on afaik so no idea.

rkuhn:
I’m afraid I lost my sonatype credentials, so I can’t publish anymore. If someone else wants to step up and needs me to confirm something, then I’ll happily do that!

Scottmitch:
@akarnokd @rkuhn - how does one become a contributor for this repo? I pinged victor offline and he is no longer involved. would love to get this merged and released.

akarnokd:
No idea. Also no idea how the release infrastructure is working.

If I had this annoying issue, after so much time, I'd seriously consider cloning the TCK, applying the fixes and begin using that to verify my projects. Given the amount of development in RS, you probably won't have to worry about becoming incompatible.

*(If I understood .NET better, I would have done it for the Reactive Streams .NET TCK repo, which practically unusable with newer test frameworks of .NET)*

ktoso (approved):
AFAIKS this looks fine, LGTM. I guess it "worked ok enough" so that we didn't notice earlier, thanks for the fix @Scottmitch.

As for releasing, I don't think I can help with that nowadays.

Scottmitch:
@ktoso - thx for review. comments addressed.

> "worked ok enough" so that we didn't notice earlier
I'm assuming most tests were "synchronous" (publisher and verification on same thread) which would make it unlikely to come up.

ktoso:
I'd approve and merge but have long since lost merge rights here. 

@viktorklang @rkuhn are you able to help out here?

rkuhn:
I can merge, so I did that, but as I said I can’t publish anymore.

ktoso:
Thank you! It seems I can publish (but not merge) huh, so I’ll give that a try if we want a patch release with this :)

rkuhn:
Sure, go ahead! I guess you’ll need another PR to bump the version, right?

Scottmitch:
version update: https://github.com/reactive-streams/reactive-streams-jvm/pull/547

anything else required for release?

PR #556, opened 2024-03-13
feat(jpms): add `module-info.java` descriptors

sgammon:
## Summary

This PR offers a minimal changeset to safely ship a `module-info.java` descriptor for the `org.reactivestreams` API artifact. The JAR is shipped as an [MRJAR](https://openjdk.org/jeps/238), with a `module-info.class` descriptor located in `META-INF/versions/9`.

This keeps the JAR compatible with JDK 6, while supporting full modularity for newer versions of Java.

Based on the current structure of the codebase, I tried to keep this PR as minimal as possible. Unfortunately, Gradle 6 does not yet have the `release` flag for the Java compiler, so I've upgraded Gradle to 7.x. The latest version is `7.6.4`. Other than this and the `module-info.java` itself, I've added a task to validate the multi-release JAR as a module.

The tests pass and everything seems fine from local, of course, but this should probably be tested with downstream artifacts. I can put together a small integration test harness which uses these JPMS-enabled artifacts with popular Reactive Streams projects to make sure there is a clear picture of impacts.

Fixes and closes #531

### Build tooling

The Gradle upgrade clears the way for some other changes, which are more optional/unrelated; for example, using [Gradle Toolchains](https://docs.gradle.org/current/userguide/toolchains.html) to enforce JDK 6 compatibility. I'm not sure if the Reactive Stream authors/maintainers want these changes, so I have split them off into another PR, which is a superset of this one: #557

Both PRs enable JPMS support, and only one or the other should be merged (if any). That PR has its own explanation about the changes enclosed.

### JAR Structure

After applying this PR, the **JAR structure** for the `api` artifact is:

Screenshot 2024-03-12 at 6 26 01 PM

> Care has been taken not to duplicate classes in the Java 9 class root, and to preserve the `FlowAdapters`.

**JAR manifest:**

Screenshot 2024-03-12 at 6 25 19 PM

**OSGi manifest (Java 9+):**

Screenshot 2024-03-12 at 6 26 34 PM

## Changelog

- feat(jpms): add `module-info.java` to `api`
- chore: adjust builds where needed to release as MRJAR artifacts
- chore: adjust BND tools to not interfere with MRJAR classes
- chore: upgrade gradle → `7.6.4`
- chore: apply updates to groovy dsl for publishing with gradle v7

PR #557, opened 2024-03-13
feat(jpms): add module-info.java descriptors (w/build upgrades)

### LINKED ISSUE #531: Consider adding a full Java module descriptor
Full Java module descriptors (`module-info.class`) are a hard requirement for building custom Java Runtimes with Jlink.

It looks like the build must remain pre Java modules (Java 6? I suppose that's related to the lowest common denominator: Android), however given that the build is Gradle based there's the https://plugins.gradle.org/plugin/org.beryx.jar plugin that can be applied to keep the build in its current Java version while at the same time providing a full module descriptor. Caveat: the generated JAR will become a multi-release JAR.

If there's any interest I'll gladly provide a PR and help with maintenance.

sgammon:
## Summary

> This intro is basically the same as #556 

This PR offers a changes to safely ship a `module-info.java` descriptor for the `org.reactivestreams` API artifact. The JAR is shipped as an [MRJAR](https://openjdk.org/jeps/238), with a `module-info.class` descriptor located in `META-INF/versions/9`.

This keeps the JAR compatible with JDK 6, while supporting full modularity for newer versions of Java.

Based on the current structure of the codebase, I tried to keep this PR as minimal as possible. Unfortunately, Gradle 6 does not yet have the `release` flag for the Java compiler, so I've upgraded Gradle to 7.x. The latest version is `7.6.4`. Other than this and the `module-info.java` itself, I've added a task to validate the multi-release JAR as a module.

The tests pass and everything seems fine from local, of course, but this should probably be tested with downstream artifacts. I can put together a small integration test harness which uses these JPMS-enabled artifacts with popular Reactive Streams projects to make sure there is a clear picture of impacts.

Fixes and closes #531

### Build Tooling

As described in the [other PR](#556), upgrading Gradle allows for use of [Gradle Toolchains](https://docs.gradle.org/current/userguide/toolchains.html) and other newer Gradle features. Conveniently, this means the build can safely be run under any JDK which supports all the targets needed by this library, and then lower bytecode levels can be requested for testing.

This PR applies those changes: the base JAR build _always_ happens against JDK 11 regardless of the toolchain used to run Gradle. This "baseline toolchain" can be overridden with, say, `./gradlew -PjavaBaseline=17`.

Given this change, I have taken other small liberties to clean up the build. For example, CI just provisions and uses JVM 21, relying on Gradle Toolchains for the rest.

### Dependency Security

I've enabled Gradle [Dependency Locking](https://docs.gradle.org/current/userguide/dependency_locking.html) and [Dependency Verification](https://docs.gradle.org/current/userguide/dependency_verification.html). This can help align build dependencies and prevent supply chain attacks (or just detect bad downloads of JARs).

I've kept these separate in commit log so they can be removed if the maintainers/authors would prefer not to activate these features. On the other hand, if this is seen as a _good_ thing, I'm also happy to take it further to enable [SLSA](https://slsa.dev/), [Sigstore](https://sigstore.dev/), and [SPDX](https://spdx.dev/), which would bring Reactive Streams up to the state of the art in dependency security.

### JAR Structure

After applying this PR, the **JAR structure** for the `api` artifact is:

Screenshot 2024-03-12 at 6 26 01 PM

> Care has been taken not to duplicate classes in the Java 9 class root, and to preserve the `FlowAdapters`.

**JAR manifest:**

Screenshot 2024-03-12 at 6 25 19 PM

**OSGi manifest (Java 9+):**

Screenshot 2024-03-12 at 6 26 34 PM

## Changelog

- feat(jpms): add `module-info.java` to `api`
- chore: adjust builds where needed to release as MRJAR artifacts
- chore: adjust BND tools to not interfere with MRJAR classes
- chore: upgrade gradle → `8.6`
- chore: apply updates to groovy dsl for publishing with gradle v8
- chore: add jdk 21 to ci
- chore: build refactors + cleanup for gradle `8.x`
- chore: enable gradle dependency locking
- chore: generate lockfiles
- chore: generate dependency verification metadata
- chore: update ci to run at latest lts


Sineaggi:
@ktoso @rkuhn Could somebody review this mr?

vietj:
what's actually blocking this ?

rkuhn:
In my case: I left the JVM ecosystem nearly ten years ago and can’t help with this anymore.

Issue #559, opened 2024-10-09
TCK requires immediate publishing of onError, but spec does not say this is required


gmarcosb:
In various TCK tests, there is an expectation that a closed/errored stream be immediately reported as such to the subscriber - see, for example, this test: https://github.com/reactive-streams/reactive-streams-jvm/blob/a625d3aba756e9842ad1291a5b73f5db280b6168/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java#L533

However, the spec does not specify whether a subscriber should immediately get these callbacks, or if it should get them after a call to `Subscription.request()`

Getting no callbacks until a request is a bit more consistent with the other stream callback behaviors

But, at a minimum, it seems like the TCK test should support the case where a `Publisher` does not call `onError/onComplete` until the `Subscriber` makes a `Subscription.request`

PR #560, opened 2024-10-09
Add a request for publishers that don't publish an error until requests

gmarcosb:


gmarcosb:
A suggested fix for https://github.com/reactive-streams/reactive-streams-jvm/issues/559

